PR #10, opened 2019-01-04
Add `description` field to OpenSeaAsset type

SirensOfTitan:
`description` seems to have been missing in the SDK.  This ensures the SDK has feature-parity with the OpenSea API.  

From what I remember a 721 can have a JSON schema defining fields other than just name, so let me know if I ought to do anything else special to include any other fields based on how your API is designed internally.  

I could not find definitive documents on the design of the OpenSea API, so I couldn't determine whether `description` is nullable or not.  Please advise here, happy to update!

## Test Plan

1. Built the project through `npm run build`, observed no errors.
2. Ran `npm test` and ensured all tests are green.
3. `yarn link`ed the updated package into our private project and spot checked that description, before this change, was null, and after applying this change is the correct value for a number of different collectibles.


alexanderatallah:
@SirensOfTitan looks great! Merging it now, and will likely deploy in version 0.4.5 this weekend

PR #17, opened 2019-07-29
Fungible token support, purchase-for, refactoring

alexanderatallah:
added "## Migrating to version 0.6" in readme with breaking changes.
Summary of changes:

1. moved tests into separate files
2. enabled VSCode type-checking for tests
3. Added 1155 support and schema
4. Added ERC20 support and schema
5. Added staticCall checking for tx.origin, DCL, and cheezewizards, all on Rinkeby only.
6. Moved all assets to a single Asset type
7. Re-did method for checking ownership of assets and wrote tests for checking and transferring quantities of them
8. Added wrapping, liquidating, and unwrapping NFTs into ERC20s on Uniswap
9. Added purchase-for functionality so you can buy an item for someone else (transferring it in one step)


src/contracts.ts
```diff
@@ -1,11 +1,16 @@
 import { PartialReadonlyContractAbi } from './types'
+import { AbiType, MethodAbi, EventAbi } from 'web3'
 
-export const getMethod = (abi: PartialReadonlyContractAbi, name: string) => {
-  return abi.filter(x => x.type == 'function' && x.name == name)[0]
+export const getMethod = (abi: PartialReadonlyContractAbi, name: string): MethodAbi => {
+  // Have to cast since there's a bug in
+  // web3 types on the 'type' field
+  return abi.filter(x => x.type == AbiType.Function && x.name == name)[0] as MethodAbi
```
joshuawu: use `find(...)`, not `filter(...)[0}`

src/contracts.ts
```diff
@@ -1,11 +1,16 @@
...
 }
 
-export const event = (abi: PartialReadonlyContractAbi, name: string) => {
-  return abi.filter(x => x.type == 'event' && x.name == name)[0]
+export const event = (abi: PartialReadonlyContractAbi, name: string): EventAbi => {
+  // Have to cast since there's a bug in
+  // web3 types on the 'type' field
+  return abi.filter(x => x.type == AbiType.Event && x.name == name)[0] as EventAbi
```
joshuawu: same

src/contracts.ts
```diff
@@ -1,11 +1,16 @@
 import { PartialReadonlyContractAbi } from './types'
+import { AbiType, MethodAbi, EventAbi } from 'web3'
 
-export const getMethod = (abi: PartialReadonlyContractAbi, name: string) => {
-  return abi.filter(x => x.type == 'function' && x.name == name)[0]
+export const getMethod = (abi: PartialReadonlyContractAbi, name: string): MethodAbi => {
+  // Have to cast since there's a bug in
```
joshuawu: It's not a bug. It's impossible to tell from the definition of `PartialReadonlyContractAbi` that the type should resolve to `MethodAbi`. To fix you need to properly type `PartialReadonlyContractAbi`.

src/utils.ts
```diff
@@ -699,38 +633,62 @@ export function estimateCurrentPrice(order: Order, secondsToBacktrack = 30, shou
...
+ * Wrapper function for getting generic Wyvern assets from OpenSea assets
+ * @param schema Wyvern schema for the asset
+ * @param asset The fungible or nonfungible asset to format
+ */
+export function getWyvernAsset(
+    schema: Schema<WyvernAsset>,
+    asset: Asset,
+    quantity = new BigNumber(1)
+  ) {
+  if (SCHEMA_NAME_TO_ASSET_CONTRACT_TYPE[schema.name as WyvernSchemaName] == AssetContractType.NonFungible) {
```
joshuawu: All these type casting is happening because `Schema` doesn't explicitly express polymorphism on `.name`. I think we should rethink the typing.

src/seaport.ts
```diff
@@ -1029,12 +1029,13 @@ export class OpenSeaPort {
       from = proxyAddress
     }
 
-    try {
+    const data = encodeTransferCall(abi, fromAddress, toAddress)
```
joshuawu: Why did you move this here?

src/seaport.ts
```diff
@@ -1136,10 +1136,11 @@ export class OpenSeaPort {
     this._dispatch(EventType.TransferOne, { accountAddress: fromAddress, toAddress, asset: wyAsset })
 
     const gasPrice = await this._computeGasPrice()
+    const data = encodeTransferCall(abi, fromAddress, toAddress)
```
joshuawu: also here. I thought it was fine on line 1142.

src/contracts.ts
alexanderatallah: changing to use `find` but there's still this annoying thing due to a bad internal lib i think:
Screen Shot 2019-08-15 at 1 50 26 PM

src/index.ts
```diff
@@ -2,6 +2,7 @@ import { OpenSeaPort } from './seaport'
 import { OpenSeaAPI } from './api'
 import { Network, EventData, EventType } from './types'
 import { orderToJSON, orderFromJSON } from './utils'
+import { encodeCall } from 'wyvern-schemas'
```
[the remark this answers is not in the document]
alexanderatallah: done

src/utils.ts
alexanderatallah: it's because the internal wyvern-schemas lib doesn't have an enum for the different schema names available. opensea-js provides that. one day we should merge the two repos

src/seaport.ts
alexanderatallah: I realized that if `encodeTransferCall` throws, then it shouldn't suppress the error. if `estimateGas` throws then the asset is not transferrable, and we should return false. but if `encodeTransferCall` throws then it's probably a developer error

src/contracts.ts
joshuawu: @alexanderatallah want to make sure this suggestion didn't get lost

PR #20, opened 2019-08-04
Implemented staticCall in Seaport.ts

michaelKim4736:
-Created individual contracts for each of the three staticCall use cases. Did we want to combine these together, or use them separately? We can combine these together if preferred, but I envisioned an individual contract for each use case, so that further use cases could be added without disruption.
-Forked .wyvern-ethereum, created a successful test case for StaticCheckTxOrigin (it's in a separate repo)
-Created _getStaticCallTargetAndExtraData function in seaport.ts. This function determines if the asset is a CheezeWizard or a DecentralandEstate. If so, it fetches the fingerprint and uses staticCall. Did we want this logic in seaport, or should the logic be elsewhere and seaport would simply receive staticExtraData and staticTarget? If seaport were to receive staticExtraData and staticTarget from elsewhere, which parameters would it be passed in through, and which functions would need to implement adding staticExtraData and staticTarget to the order? I updated _makeSellOrder() and _makeBuyOrder(), since those seemed to be the core functions re-used by others, and used the _getStaticCallTargetAndExtraData() function that I created.
-Currently missing some addresses for contracts in utils.ts. For example, Dapper has not yet deployed the mainnet version of the main CheezeWizard contract, I don't know which OpenSeaAdmin addresses we want to use, Decentraland uses Ropstein not Rinkeby, and I have not yet deployed the StaticCall contracts since I don't yet know whether we want to combine them into a single contract or not.


src/seaport.ts
```diff
@@ -1746,8 +1809,17 @@ export class OpenSeaPort {
         ? makeBigNumber(sellOrder.takerRelayerFee)
         : makeBigNumber(sellOrder.makerRelayerFee)
     } else {
+      // If all assets are for the same contract and it's a non-private sale, use its fees
+      let asset: OpenSeaAsset | null = null
+      if (assets && _.uniqBy(assets, a => a.tokenAddress).length == 1) {
+        const { tokenAddress, tokenId } = assets[0]
+        asset = await this.api.getAsset(tokenAddress, tokenId)
```
alexanderatallah: let's change the type of getAsset to OpenSeaAsset (nonnull), and remove the json null check there. that way you won't have to check for null all over the place here. 404s are thrown as errors by Transport.ts

src/seaport.ts
```diff
@@ -1244,23 +1254,21 @@ export class OpenSeaPort {
...
    * @param assetContract Optional prefetched asset contract (including fees) to use instead of assets
    * @param side The side of the order (buy or sell)
    * @param isPrivate Whether the order is private or not (known taker)
    * @param extraBountyBasisPoints The basis points to add for the bounty. Will throw if it exceeds the assets' contract's OpenSea fee.
    */
   public async computeFees(
-      { assets, assetContract, side, isPrivate = false, extraBountyBasisPoints = 0 }:
-      { assets?: Asset[];
+      { asset, assetContract, side, isPrivate = false, extraBountyBasisPoints = 0 }:
+      { asset: OpenSeaAsset | null;
```
alexanderatallah: we should just use `OpenSeaAsset?` so that we don't need `asset: null` everywhere

src/seaport.ts
```diff
@@ -1653,37 +1659,42 @@ export class OpenSeaPort {
...
     ): Promise<UnhashedOrder> {
 
-    accountAddress = validateAndFormatWalletAddress(this.web3, accountAddress)
-    const schema = this._getSchema(schemaName)
-    const wyAsset = getWyvernNFTAsset(schema, asset.tokenId, asset.tokenAddress)
-    const isPrivate = buyerAddress != NULL_ADDRESS
-    const { totalSellerFeeBPS,
-            totalBuyerFeeBPS,
-            sellerBountyBPS } = await this.computeFees({ assets: [asset], side: OrderSide.Sell, isPrivate, extraBountyBasisPoints })
+        accountAddress = validateAndFormatWalletAddress(this.web3, accountAddress)
```
alexanderatallah: indentation here got messed up with an extra two spaces

src/seaport.ts
```diff
@@ -1716,6 +1727,61 @@ export class OpenSeaPort {
     }
   }
 
+  public async _getStaticCallTargetAndExtraData(
+      { asset, useTxnOriginStaticCall }:
+      { asset: OpenSeaAsset;
+        useTxnOriginStaticCall: boolean; }
+    ): Promise<any> {
+        const isCheezeWizards = String(asset.tokenAddress).toLowerCase() === String(CHEEZE_WIZARDS_GUILD_ADDRESS).toLowerCase() || String(asset.tokenAddress).toLowerCase() === String(CHEEZE_WIZARDS_GUILD_RINKEBY_ADDRESS).toLowerCase()
```
alexanderatallah: indentation should be two spaces

src/seaport.ts
```diff
@@ -1716,6 +1727,61 @@ export class OpenSeaPort {
...
+        const isDecentralandEstate = String(asset.tokenAddress).toLowerCase() === String(DECENTRALAND_ESTATE_ADDRESS).toLowerCase()
+        const isMainnet = (this._networkName === Network.Main)
```
alexanderatallah: nit: unnecessary parentheses

alexanderatallah: same with `(isMainnet)` below

test/seaport.ts
```diff
@@ -558,6 +551,54 @@ suite('seaport', () => {
 
   test("Matches a private sell order, doesn't for wrong taker", async () => {
     const accountAddress = ALEX_ADDRESS
+    console.log(ALEX_ADDRESS)
```
alexanderatallah: remove log

test/seaport.ts
```diff
@@ -558,6 +551,54 @@ suite('seaport', () => {
...
+    const takerAddress = ALEX_ADDRESS_2
+    const amountInToken = 2
+    const bountyPercent = 0
+
+    const tokenId = MYTHEREUM_TOKEN_ID.toString()
+    const tokenAddress = MYTHEREUM_ADDRESS
+
+    const asset = await client.api.getAsset(tokenAddress, tokenId)
+    assert.isNotNull(asset)
```
alexanderatallah: no longer necessary! same with if (!asset) ...

test/seaport.ts
```diff
@@ -558,6 +551,54 @@ suite('seaport', () => {
...
+    try {
+      await testMatchingNewOrder(order, DEVIN_ADDRESS)
+    } catch (e) {
+      // It works!
+      return
+    }
+    assert.fail()
+  })
+
+  test.only("Mainnet StaticCall Tx.Origin: Matches a private sell order, doesn't for wrong taker", async () => {
```
alexanderatallah: let's remove the private sell order part since that's a separate safeguard tested above. buyerAddress below should be NULL_ADDRESS

test/seaport.ts
```diff
@@ -571,6 +612,53 @@ suite('seaport', () => {
...
+      accountAddress,
+      startAmount: amountInToken,
+      extraBountyBasisPoints: bountyPercent * 100,
+      buyerAddress: takerAddress,
+      expirationTime: 0,
+      paymentTokenAddress: NULL_ADDRESS,
+      waitForHighestBid: false,
+      schemaName: WyvernSchemaName.ERC721
+    })
+    console.log(order);
```
alexanderatallah: remove log

test/seaport.ts
```diff
@@ -602,6 +690,50 @@ suite('seaport', () => {
...
+    const tokenAddress = '0x095731b672b76b00A0b5cb9D8258CD3F6E976cB2' // Testnet CheezeWizards Guild address
+
+    const asset = await rinkebyClient.api.getAsset(tokenAddress, tokenId)
+
+    const order = await rinkebyClient._makeSellOrder({
+      asset: { tokenAddress, tokenId },
+      accountAddress,
+      startAmount: amountInToken,
+      extraBountyBasisPoints: bountyPercent * 100,
+      buyerAddress: takerAddress,
```
alexanderatallah: same here

src/seaport.ts
michaelKim4736: We sometimes pass in null in _makeBundleBuyOrder and _makeBundleSellOrder, if the bundle is not all from the same contract

alexanderatallah:
Closing in favor of https://github.com/ProjectOpenSea/opensea-js/pull/26

PR #31, opened 2020-02-07
feat(add-last-sale-type): Add LastSale type and data augmentation methods

taylorjdawson:


src/types.ts
```diff
@@ -385,12 +389,33 @@ export interface OpenSeaAsset extends Asset {
...
 
+export interface LastSale {
+  eventType: string
+  auctionType: string
+  totalPrice: string
+  transaction: Transaction
+}
+
+export interface Transaction {
+  id: number
```
alexanderatallah: what's `id` used for? except for Account pks (to help with multiple blockchains in the future), we shouldn't need the pks of any other model in frontend serializers

src/types.ts
```diff
@@ -385,12 +389,33 @@ export interface OpenSeaAsset extends Asset {
...
+  timestamp: string
+}
+
 /**
  * Full annotated Fungible Token spec with OpenSea metadata
  */
 export interface OpenSeaFungibleToken extends Token {
   imageUrl?: string
   ethPrice?: string
+  usdPrice?: string
```
alexanderatallah: don't see the serialization code for this line

src/utils.ts
```diff
@@ -255,6 +291,31 @@ export const assetFromJSON = (asset: any): OpenSeaAsset => {
   return fromJSON
 }
 
+export const transactionFromJSON = (transaction: any): Transaction => {
+  return {
+    id: transaction.id,
+    fromAccount: accountFromJSON(transaction.from_account),
+    toAccount: accountFromJSON(transaction.to_account),
+    createdDate: transaction.created_date,
```
alexanderatallah: we've been converting these to javascript Dates in other places, so let's do that here too? `new Date(`${transaction.created_date}Z`)` and same for modifiedDate

src/types.ts
taylorjdawson: Good catch!

src/utils.ts
taylorjdawson: 👍

src/types.ts
taylorjdawson: What is "Account pks"?

alexanderatallah: The ids for Account models (OpenSeaAccount in this file). We can’t always identify an account by its address alone, so we’ll need the id

src/types.ts
```diff
@@ -389,24 +389,54 @@ export interface OpenSeaAsset extends Asset {
...
 
+/**
+ * Defines a Transaction type.
+ */
 export interface Transaction {
-  id: number
+  // The details about the account that sent the transaction
   fromAccount: OpenSeaAccount
+
+  // The details about the account that received the transaction
```
alexanderatallah: thanks for adding these!

src/types.ts
```diff
@@ -389,24 +389,54 @@ export interface OpenSeaAsset extends Asset {
   transferFeePaymentToken: OpenSeaFungibleToken | null
 }
 
+/**
+ * Defines a LastSale type which contains details about last sale of the token
+ */
 export interface LastSale {
```
alexanderatallah: thought we were gonna add event_timestamp?

src/types.ts
```diff
@@ -389,24 +389,54 @@ export interface OpenSeaAsset extends Asset {
...
+
+  // The type of event
   eventType: string
+
+  // The auction type
   auctionType: string
+
+  // The total price of the sale in the payment
```
alexanderatallah: just reminded me that this price is in the event's payment_token, which isn't being serialized here (we need to do that)

taylorjdawson: `totalPrice` does get serialized [here](https://github.com/ProjectOpenSea/opensea-js/pull/31/files#diff-e1495d267619047a7cca5cfe8f692729R272)

.nvmrc
```diff
@@ -1 +1 @@
-8.11.2
\ No newline at end of file
+10.19.0
```
taylorjdawson: @alexanderatallah updated the node version since node v8 is now outdated

src/types.ts
```diff
@@ -207,8 +207,12 @@ export interface OpenSeaAccount {
   address: string
   // Public configuration info, including "affiliate" for users who are in the OpenSea affiliate program
   config: string
+
+  // TODO: Add description
+  discordId: string
+
   // This account's profile image - by default, randomly generated by the server
-  profile_img_url: string
+  profileImgUrl: string
```
taylorjdawson: Changed this to camel case to match the rest of the variables

src/types.ts
```diff
@@ -385,12 +389,66 @@ export interface OpenSeaAsset extends Asset {
...
+  eventTimestamp: Date
+
+  // The auction type
+  auctionType: string
+
+  // The total price of the sale in the payment
+  totalPrice: string
+
+  // The transaction associated with the token sale
+  transaction?: Transaction
```
taylorjdawson: Not sure if we want to do `transaction: Transaction | null` here instead?

src/utils.ts
```diff
@@ -999,7 +1061,7 @@ export async function getNonCompliantApprovalAddress(erc721Contract: Web3.Contra
     // CRYPTOKITTIES check
     promisifyCall<string>(c => erc721Contract.kittyIndexToApproved.call(tokenId, c), onError),
     // Etherbots check
-    promisifyCall<string>(c => erc721Contract.partIndexToApproved.call(tokenId, c), onError),
+    // promisifyCall<string>(c => erc721Contract.partIndexToApproved.call(tokenId, c), onError),
```
taylorjdawson: @alexanderatallah I commented this out since I did not understand it's usage as if you look in the tests it passes a reference of the Cryptokittes contract instances to this method; `erc721Contract` is the CK contract. `erc721Contract.partIndexToApproved` should fail because `partIndexToApproved` is on the Etherbots contract AFAIK.

Please let me know if these assumptions are true so that I can then remove the commented out line

src/utils.ts
```diff
@@ -266,12 +266,13 @@ export const assetFromJSON = (asset: any): OpenSeaAsset => {
     openseaLink: asset.permalink,
     traits: asset.traits,
     numSales: asset.num_sales,
-    lastSale: {
+    lastSale: asset.last_sale ? {
```
alexanderatallah: can we make a helper called `assetEventFromJSON` to do this object building, consistent with the other ones above?

src/utils.ts
alexanderatallah: @taylorjdawson I think we need to keep this and fix the onError handler, otherwise I could write an etherbots test that would fail (this is currently failing on a cryptokitties test, but there is no etherbots test).

Also, the line `let approvedAddr = await promisifyCall<string>(c => erc721.getApproved.call(tokenId, c))` in `public async approveNonFungibleToken` would also have problems

alexanderatallah: Let's just remove onError and make that argument undefined. don't think it's needed - I just wasn't getting errors with geth and added it to make sure we throw if we haven't updated the erc721 abi to include weird approval methods, but we have done that

src/types.ts
alexanderatallah: yeah let's do that and make it nullable

.nvmrc
alexanderatallah: @taylorjdawson sounds good if it works now. there used to be a node-scrypt issue with 10 and 12. (also might be worth trying 12)

src/types.ts
alexanderatallah: maybe the line moved - it links to ` auctionType: asset.last_sale.auction_type,` which serializes whether it was an english/dutch auction. By payment_token, i mean we need to call `tokenFromJSON` on the asset_event.payment_token, since total_price is in that payment token, and if you don't know if it's eth or some erc20 token, you don't really know the price at all

taylorjdawson:
> added comments. we'll need to test this inside of the frontend too, since serializers were modified. I can show you how to do that easily on Monday!

Let's find a time to do this

src/types.ts
diklou: What does it mean when transaction is null ? Does it mean it's pending ?

michcio1234: @taylorjdawson @alexanderatallah @diklou 
>What does it mean when transaction is null ? Does it mean it's pending ?

up ⬆️

PR #33, opened 2020-02-25
taylor/fix(remove-discord-id): 

taylorjdawson:
Removes the `discordId` from `OpenSeaAccount`
 - discordId field will be added to the `user` attribute in the future so leaving it out for now


alexanderatallah:
You didn't add the bignumber import in constants.ts, so what's the fix here? the code tree also still has the issue: https://github.com/ProjectOpenSea/opensea-js/blob/754552873195bbcb7f4b57320cf3fe8263aa9277/lib/constants.d.ts#L9

PR #34, opened 2020-02-26
Taylor/feat/add user serializer

taylorjdawson:


src/types.ts
```diff
@@ -215,10 +215,12 @@ export interface OpenSeaAccount {
   profileImgUrl: string
 
   // More information explicitly set by this account's owner on OpenSea
-  user: null | {
-    // Username for this account
-    username: string
-  }
+  user: null | OpenSeaUser
```
alexanderatallah: nit: usually we flip these to put null at the end (sorta consistent with `?`)

PR #35, opened 2020-03-02
Multi-schema bundle support

alexanderatallah:
This is the start of support for 1155 bundles. Involves:

Refactoring:
- moving atomicize and encoding logic out of wyvern-schemas and into opensea-js
- consolidating functions like `getWyvernNFTAsset` into one
- removed some deprecated code from previous minor versions
- moving schemaName into the Asset type, which allows...

New features
- enabling bundling for fungible and semifungible items
- allowing bundles to contain assets with more than one schema

Compat:
- backwards compat for bundles with one schema listed

Tests:
- refactoring tests and utils into more files
- added tests for multi-schema bundles and fungible bundles

Also reduced the size of wyvern-schemas significantly by removing unused schemas.


.gitignore
```diff
@@ -62,3 +62,5 @@ typings/
 
 # next.js build output
 .next
+
+.vscode
```
joshuawu: shouldn't this belong in your `.git/info/exclude`?

src/utils/schema.ts
```diff
@@ -0,0 +1,234 @@
+import { BigNumber } from 'bignumber.js'
```
joshuawu: Which files had their content moved here?

alexanderatallah: this file is mostly new (took methods we wrote for atomicizer logic out of wyvern-schemas), aside from `encodeAtomicizedTransfer` and `encodeTransferCall`, which were moved out of utils.ts

.gitignore
alexanderatallah: fixed!

README.md
```diff
@@ -90,9 +91,43 @@ const seaport = new OpenSeaPort(provider, {
...
+/**
+ * Simple, unannotated non-fungible asset spec
+ */
+export interface Asset {
+  // The asset's token ID, or null if ERC-20
+  tokenId: string | null,
+  // The asset's contract address
+  tokenAddress: string,
+  // The Wyvern schema name (defaults to "ERC721") for this asset
+  schemaName?: WyvernSchemaName,
```
taylorjdawson: Is this field optional? Or should it be non-nullable

package.json
```diff
@@ -40,28 +40,29 @@
...
   "dependencies": {
     "bignumber.js": "^4.0.4",
+    "ethereumjs-abi": "git+https://github.com/ProjectWyvern/ethereumjs-abi.git",
     "ethereumjs-util": "^5.1.3",
     "fbemitter": "^2.1.1",
     "isomorphic-unfetch": "^2.1.1",
     "json-loader": "^0.5.7",
     "lodash": "^4.17.4",
-    "query-string": "^6.1.0",
+    "query-string": "6.11.0",
```
taylorjdawson: minor but should we keep the carrots `^` so packages will get minor/patch versions automatically?

package.json
```diff
@@ -40,28 +40,29 @@
...
     "webpack": "^3.10.0",
-    "wyvern-js": "github:ProjectOpenSea/wyvern-js#semver:^3.2.0",
-    "wyvern-schemas": "github:ProjectOpenSea/wyvern-schemas#semver:^0.10.7"
+    "wyvern-js": "github:ProjectOpenSea/wyvern-js#semver:^3.2.1",
+    "wyvern-schemas": "github:ProjectOpenSea/wyvern-schemas#semver:^0.11.1"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run lint && npm test && npm run build"
+      "pre-commit": "npm run lint && npm run build && git add ."
```
taylorjdawson: It would be better to do it this way:
```
  "husky": {
    "hooks": {
      "pre-commit": "npm run lint && npm test && npm run build",
      "post-commit": "git add lib/*"
    }
  },
```
`git add lib/*` won't add all the files in the directory as there could be times we don't want to commit all files.
Also it will only run `git add lib/*` if you complete the commit.

src/seaport.ts
```diff
@@ -458,40 +465,29 @@ export class OpenSeaPort {
...
     ): Promise<Order> {
 
-    if (!assets && tokenIds && tokenAddresses) {
-      onDeprecated("Use `assets` instead of the `tokenIDs` and `tokenAddresses` list")
-      assets = _.zipWith(tokenIds, tokenAddresses, (tokenId, tokenAddress) => {
-        return { tokenAddress, tokenId }
-      })
-    }
-
+    quantities = quantities || assets.map(a => 1)
```
taylorjdawson: "Magic Number"
Might be good to comment here as this could be mysterious

taylorjdawson: Better yet add `quantities` to the js doc comment and say it defaults to 1 like the `createBuyOrder` method below does

src/seaport.ts
```diff
@@ -834,27 +808,30 @@ export class OpenSeaPort {
    * @param schemaName The Wyvern schema name corresponding to the asset type
    */
   public async createBundleSellOrder(
-      { bundleName, bundleDescription, bundleExternalLink, assets, accountAddress, startAmount, endAmount, expirationTime = 0, waitForHighestBid = false, paymentTokenAddress, extraBountyBasisPoints = 0, buyerAddress, schemaName = WyvernSchemaName.ERC721 }:
+      { bundleName, bundleDescription, bundleExternalLink, assets, quantities, accountAddress, startAmount, endAmount, expirationTime = 0, waitForHighestBid = false, paymentTokenAddress, extraBountyBasisPoints = 0, buyerAddress }:
       { bundleName: string;
         bundleDescription?: string;
         bundleExternalLink?: string;
         assets: Asset[];
+        quantities?: number[];
```
taylorjdawson: nit: add to js docs

src/seaport.ts
```diff
@@ -834,27 +808,30 @@ export class OpenSeaPort {
    * @param schemaName The Wyvern schema name corresponding to the asset type
```
taylorjdawson: nit: remove from js docs

src/seaport.ts
```diff
@@ -2273,8 +2199,9 @@ export class OpenSeaPort {
       salt: WyvernProtocol.generatePseudoRandomSalt(),
       metadata: {
         bundle,
-        schema: schema.name as WyvernSchemaName,
-      },
+        // TODO deprecate
+        schema: schemas[0].name as WyvernSchemaName
```
taylorjdawson: should we null check here?

package.json
alexanderatallah: good catch. we might want to make "~" the default because we've run into people doing breaking changes in their minor versions before. will add

src/seaport.ts
alexanderatallah: done

src/seaport.ts
alexanderatallah: no need, it's deprecated and checked earlier (there's a check that you can't have fewer than one asset/schema in getWyvernAsset)

PR #36, opened 2020-03-19
Update prices to be BigNumbers

taylorjdawson:


lib/constants.d.ts
```diff
@@ -1,13 +1,12 @@
-import { BigNumber } from 'bignumber.js';
```
alexanderatallah: need this to fix the weird import error

lib/constants.d.ts
```diff
@@ -1,13 +1,12 @@
...
 export declare const DEFAULT_GAS_INCREASE_FACTOR = 1.1;
-export declare const NULL_ADDRESS: string;
+export declare const NULL_ADDRESS: any;
```
alexanderatallah: how did this happen? looks like something changed - npm / typescript version?

src/types.ts
```diff
@@ -462,8 +462,8 @@ export interface Transaction {
  */
 export interface OpenSeaFungibleToken extends Token {
   imageUrl?: string
-  ethPrice?: string
-  usdPrice?: string
+  ethPrice?: BigNumber
```
alexanderatallah: what about the corresponding `utils` deserializers for these types? strings need to be (safely) turned into bignumbers using `makeBigNumber`

src/utils/utils.ts
```diff
@@ -461,7 +461,7 @@ export const orderToJSON = (order: Order): OrderJSON => {
     staticExtradata: order.staticExtradata,
     paymentToken: order.paymentToken.toLowerCase(),
     quantity: order.quantity.toString(),
-    basePrice: order.basePrice.toString(),
+    basePrice: makeBigNumber(order.basePrice),
```
alexanderatallah, line 464: how did this typecheck? the orderToJSON method needs to output valid JSON, which needs to convert bignumbers to strings.

package-lock.json
```diff
@@ -3560,13 +3568,13 @@
...
         "wrappy": {
           "version": "1.0.2",
-          "resolved": false,
+          "resolved": "https://registry.npmjs.org/wrappy/-/wrappy-1.0.2.tgz",
           "integrity": "sha1-tSQ9jz7BqjXxNkYFvA0QNuMKtp8=",
           "optional": true
         },
         "yallist": {
           "version": "3.1.1",
-          "resolved": false,
```
alexanderatallah, line 3569: so odd that this sometimes turns to `false`. I might have something configured wrong. I googled this problem and only found old irrelevant comments about npm bugs. lmk if you know

package-lock.json
```diff
@@ -7353,7 +7361,8 @@
         "axios": "^0.17.1",
         "bignumber.js": "^6.0.0",
         "typed-promisify": "^0.4.0",
-        "web3-provider-engine": "^13.6.5"
+        "web3-provider-engine": "^13.6.5",
+        "wyvern-js": "github:ProjectOpenSea/wyvern-js#semver:^3.2.1"
       },
```
alexanderatallah, line 7366: woah how did this get added back? we don't need the `wyvern-js` dependency anymore

taylorjdawson: @alexanderatallah possibly my package-lock.json

PR #40, opened 2020-04-23
Points node url to opensea's json rpc proxy

taylorjdawson:


src/constants.ts
```diff
@@ -6,8 +6,8 @@ export const NULL_ADDRESS = WyvernProtocol.NULL_ADDRESS
 export const NULL_BLOCK_HASH = '0x0000000000000000000000000000000000000000000000000000000000000000'
 export const OPENSEA_FEE_RECIPIENT = '0x5b3256965e7c3cf26e11fcaf296dfc8807c01073'
 export const DEP_INFURA_KEY = 'e8695bce67944848aa95459fac052f8e'
-export const MAINNET_PROVIDER_URL = 'https://eth-mainnet.alchemyapi.io/jsonrpc/y5dLONzfAJh-oCY02DCP3UWCT2pSEXMo'
-export const RINKEBY_PROVIDER_URL = 'https://eth-rinkeby.alchemyapi.io/jsonrpc/-yDg7wmgGw5LdsP4p4kyxRYuDzCkXtoI'
+export const MAINNET_PROVIDER_URL = ' https://api.opensea.io/jsonrpc/v1'
```
alexanderatallah, line 9: leading space. also, can we add an appending slash? Django recommends it, and adding appending slashes has fixed mysterious occasional issues with some browsers. https://stackoverflow.com/questions/42212122/why-django-urls-end-with-a-slash/42213107

taylorjdawson:
@alexanderatallah Ran through the following tests on rinkeby:
- posting sell order of bundle
- posting sell order of single order
- posting a bid
- buying an asset

Sound good?

alexanderatallah:
@taylorjdawson sounds good!

PR #43, opened 2020-06-25
Remove gasPrice in txnData for estimating gas

taylorjdawson:


src/seaport.ts
```diff
@@ -1828,8 +1828,7 @@ export class OpenSeaPort {
     this._dispatch(EventType.InitializeAccount, { accountAddress })
     this.logger(`Initializing proxy for account: ${accountAddress}`)
 
-    const gasPrice = await this._computeGasPrice()
-    const txnData: any = { from: accountAddress, gasPrice }
+    const txnData: any = { from: accountAddress }
```
dfinzer, line 1831: Hmm but then we've removed the gas price when we submit the txn right? We need to pass in a different dict in `estimateGasAsync`

taylorjdawson: @dfinzer ah right, fixed

PR #52, opened 2020-11-10
stab at adding reserve price to the sdk

DJViau:


src/seaport.ts
```diff
@@ -598,20 +598,22 @@ export class OpenSeaPort {
...
    * @param paymentTokenAddress Address of the ERC-20 token to accept in return. If undefined or null, uses Ether.
    * @param extraBountyBasisPoints Optional basis points (1/100th of a percent) to reward someone for referring the fulfillment of this order
    * @param buyerAddress Optional address that's allowed to purchase this item. If specified, no other address will be able to take the order, unless its value is the null address.
    * @param buyerEmail Optional email of the user that's allowed to purchase this item. If specified, a user will have to verify this email before being able to take the order.
    */
   public async createSellOrder(
-      { asset, accountAddress, startAmount, endAmount, quantity = 1, expirationTime = 0, waitForHighestBid = false, paymentTokenAddress, extraBountyBasisPoints = 0, buyerAddress, buyerEmail }:
+      { asset, accountAddress, startAmount, endAmount, quantity = 1, expirationTime = 0, waitForHighestBid = false, englishAuctionReservePrice = 0, paymentTokenAddress, extraBountyBasisPoints = 0, buyerAddress, buyerEmail }:
```
alexanderatallah, line 608: nit: we can default the reserve price to zero in `_makeSellOrder` and `_makeBundleSellOrder` instead. that way it's only being done in one place

src/seaport.ts
```diff
@@ -2870,6 +2881,9 @@ export class OpenSeaPort {
     if (priceDiff > 0 && expirationTime == 0) {
       throw new Error('Expiration time must be set if order will change in price.')
     }
+    if (englishAuctionReservePrice && (startAmount < englishAuctionReservePrice)) {
+      throw new Error('Reserrve price must be greater than or equal to the start amount.')
```
alexanderatallah, line 2885: misspelling

src/seaport.ts
```diff
@@ -2017,6 +2024,7 @@ export class OpenSeaPort {
       takerProtocolFee,
       makerReferrerFee,
       waitingForBestCounterOrder: waitForHighestBid,
+      englishAuctionReservePrice: makeBigNumber(englishAuctionReservePrice || 0),
```
alexanderatallah, line 2027: since `englishAuctionReservePrice` can be undefined in the UnhashedOrder type that you defined below (which i think is better than having it defined for e.g. non-english auctions), i would have this set to `undefined` instead of `makeBigNumber(0)`

src/seaport.ts
```diff
@@ -2261,6 +2270,7 @@ export class OpenSeaPort {
...
```
alexanderatallah, line 2273: `orderToJSON` and `OrderType` also need to be updated, to send the data in our POST request

src/seaport.ts
alexanderatallah: Add a unit test in test/seaport/orders.ts similar to this one:
```
try {
      await client._makeSellOrder({
        asset: { tokenAddress, tokenId },
        quantity: 1,
        accountAddress,
        startAmount: 2,
        extraBountyBasisPoints: 0,
        buyerAddress: NULL_ADDRESS,
        expirationTime: 0,
        paymentTokenAddress,
        waitForHighestBid: true,
      })
      assert.fail()
    } catch (error) {
      assert.include(error.message, 'English auctions must have an expiration time')
    }
```

src/types.ts
```diff
@@ -567,6 +567,7 @@ export interface UnhashedOrder extends WyvernOrder {
   // OpenSea-specific
   makerReferrerFee: BigNumber
   waitingForBestCounterOrder: boolean
+  englishAuctionReservePrice?: BigNumber;
```
alexanderatallah, line 570: in `test/api/api.ts`, let's start a unit test to test that orderToJSON is doing the right thing:
```

test('orderToJSON is correct', async () => {
    const accountAddress = ALEX_ADDRESS
    const takerAddress = ALEX_ADDRESS_2
    const amountInToken = 1.2
    const paymentTokenAddress = WETH_ADDRESS
    const expirationTime = Math.round(Date.now() / 1000 + 60) // one minute from now

    const tokenId = MYTHEREUM_TOKEN_ID.toString()
    const tokenAddress = MYTHEREUM_ADDRESS
    const order = await client._makeSellOrder({
      asset: { tokenAddress, tokenId },
      quantity: 1,
      accountAddress,
      startAmount: amountInToken,
      paymentTokenAddress,
      extraBountyBasisPoints: 0,
      buyerAddress: NULL_ADDRESS,
      expirationTime,
      waitForHighestBid: true,
      englishAuctionReservePrice: 2
    })

    const orderData = orderToJSON(order)
    assert.equal(...)
  })
```

alexanderatallah (changes requested):
@DJViau added comments

src/seaport.ts
```diff
@@ -2881,8 +2881,8 @@ export class OpenSeaPort {
     if (priceDiff > 0 && expirationTime == 0) {
       throw new Error('Expiration time must be set if order will change in price.')
     }
-    if (englishAuctionReservePrice && (startAmount < englishAuctionReservePrice)) {
-      throw new Error('Reserrve price must be greater than or equal to the start amount.')
+    if (englishAuctionReservePrice && (englishAuctionReservePrice < startAmount)) {
```
alexanderatallah, line 2884: should have mentioned this earlier but we should also check that `waitingForBestCounterOrder` is true if you have a `englishAuctionReservePrice` set

alexanderatallah: also yay the tests caught a big bug?

src/seaport.ts
```diff
@@ -2881,7 +2895,7 @@ export class OpenSeaPort {
       ? makeBigNumber(this.web3.toWei(priceDiff, 'ether')).round()
       : WyvernProtocol.toBaseUnitAmount(makeBigNumber(priceDiff), token.decimals)
 
-    return { basePrice, extra, paymentToken }
+    return { basePrice, extra, paymentToken, englishAuctionReservePrice }
```
alexanderatallah, line 2898: should have caught this earlier but we should have basePrice and englishAuctionReservePrice use the same unit system, e.g. both should be converted to the base unit amount

test/api/api.ts
```diff
@@ -68,6 +80,46 @@ suite('api', () => {
...
+
+    const orderData = orderToJSON(hashedOrder)
+    assert.equal(orderData.quantity, quantity.toString())
+    assert.equal(orderData.maker, accountAddress)
+    assert.equal(orderData.taker, NULL_ADDRESS)
+    assert.equal(orderData.basePrice, WyvernProtocol.toBaseUnitAmount(makeBigNumber(amountInToken), 18).toString())
+    assert.equal(orderData.paymentToken, paymentTokenAddress)
+    assert.equal(orderData.extra, extraBountyBasisPoints.toString())
+    assert.equal(orderData.expirationTime, expirationTime + ORDER_MATCHING_LATENCY_SECONDS)
+    assert.equal(orderData.englishAuctionReservePrice, englishAuctionReservePrice.toString())
```
alexanderatallah, line 120: see note above about having it be in the base unit amount same as basePrice

alexanderatallah (commented):
left comments

alexanderatallah (approved):

PR #153, opened 2021-08-31
Remove special order handling for Private Sales

royliu2:


src/seaport.ts
```diff
@@ -1592,15 +1592,6 @@ export class OpenSeaPort {
       throw new Error(errorMessage)
     }
 
-    // Remove fees for private orders
-    if (isPrivate) {
```
alexanderatallah, line 1596: think isPrivate doesn't need to be a param to this method anymore

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 Full changelog coming soon. For now, here are the most important changes for doing major migrations:
 
+## Migrating to version 1.2
+- `OpenSeaPort::computeFees()` No longer accepts an `isPrivate` parameter.
```
royliu2, line 6: Think we should have something in the CHANGELOG for this, but not sure about new version #. Should we bump to 1.2 or 1.1.4?

sunwrobert: 1.2? i would guess

sunwrobert (approved):
lgtm personally

michael-cohen-eth (approved):

PR #196, opened 2021-10-04
Update utils.ts

git-sgmoore:
Fix typo at line 139


ryanio:
falsey is correct in this context

PR #209, opened 2021-10-15
construct RPC urls from apiBaseUrl

royliu2:
Update the Seaport constructor to use RPC urls based on the api config apiBaseUrl


Meemaw (approved):
LGTM

PR #221, opened 2021-10-24
remove inconsistent references to props

Mortaro:
Some of the examples were referring to seaport directly and on the same snippet referring to seaport from this.props. 

I removed the reference to props since this library is supposed to work with vanilla js without react.


ryanio:
thank you, have added to my PR to collect together https://github.com/ProjectOpenSea/opensea-js/pull/849

PR #265, opened 2021-12-01
MP fulfill order gas estimation

jtaaa:
Adds two public methods that MoonPay uses for the ETH direct purchase flow: `getFulfillOrderGasEstimation` and `getFulfillOrderArgs`. Because of an incompatibility between their signing library ethers.js and ours web3.js, MoonPay is unable to use our SDK directly. Instead they get the call args and gas estimation necessary and submit the transaction themselves.

## Technical Description

Extracted 3 helpers from `_atomicMatch` that:
1. Validates the order and order match
2. Gets the call args
3. Estimates gas

Use those helpers for MoonPay's public methods.

## Note

There is a similar public function `_estimateGasForMatch` that does same job as `getFulfillOrderGasEstimation` and `_getAtomicMatchGasEstimation` with retry handling but takes different params. I've left both functions in, instead of trying to merge and re-use to simplify retaining the same error handling in the original `_atomicMatch` fn.


conanbatt:
Can we add tests to this?

conanbatt:
I cant figure out the functional validity of the code, I would very much appreciate some unit tests for that.

JoshuaSchmidt-OpenSea:
This PR is out of date so I am closing without merging.

PR #270, opened 2021-12-06
Accomodate new Wyvern Fee wrapper flows

sunwrobert:
This change assumes that the Order will have the relevant serverSignature and feeData. If the order has the static target as our wyvern fee wrapper address, then we will go into the fee wrapper flow. 

The fee wrapper flow 
1. Requests a signature for the order for the person submitting the tx. This adds one extra step (signing) to every new order fulfillment in this flow
2. Then does an estimateGas/sendTransaction of the atomicMatch_ function in the fee wrapper.


src/seaport.ts
```diff
@@ -867,12 +869,38 @@ export class OpenSeaPort {
     const { buy, sell } = assignOrdersToSides(order, matchingOrder)
 
     const metadata = this._getMetadata(order, referrerAddress)
-    const transactionHash = await this._atomicMatch({ buy, sell, accountAddress, metadata })
+
+    let transactionHash: string = '';
+
+    if (isFeeWrapperStaticTarget({buy, sell})) {
```
vickygos, line 875: Can we add comments for why we're doing this and context?

vickygos: Do we need to check fee recipient is fee wrapper?

sunwrobert: hmm, we could but that should be validated already on our backend (i.e. a user should never be able to create an order with the fee recipient not being the wrapper)

src/seaport.ts
```diff
@@ -867,12 +870,40 @@ export class OpenSeaPort {
...
+        buy:
+          properlySignedMatchingOrder.side === OrderSide.Buy
+            ? properlySignedMatchingOrder
+            : buy,
+        sell:
+          properlySignedMatchingOrder.side === OrderSide.Sell
+            ? properlySignedMatchingOrder
+            : sell,
+        accountAddress,
+        metadata,
```
vickygos, line 890: how come you don't pass in `makerOrder` here?

vickygos: i dont think you need it in line  898 since `if (useFeeWrapper && makerOrder?.hash) {`
will only ever be false with useFeeWrapper = false

src/seaport.ts
```diff
@@ -2927,10 +2961,80 @@ export class OpenSeaPort {
...
     ]
 
+    const useFeeWrapper = isFeeWrapperFlow({ buy, sell }, this._networkName)
+
+    let wyvernFeeWrapperArgs: WyvernFeeWrapperAtomicMatchParameters | undefined
+
+    if (useFeeWrapper && makerOrder?.hash) {
+      const { fulfillmentData } = await this.api.getOrderFulfillmentData(makerOrder.hash)
+
+      wyvernFeeWrapperArgs = [args, fulfillmentData.serverSignature || '', fulfillmentData.feeData]
```
vickygos, line 2971: Should we throw exception/fail here is serverSignature is empty? I think that's weird

vickygos: Also for testing, wanna log if feeData is empty?

src/seaport.ts
sunwrobert: ah yeah, mistake

src/seaport.ts
sunwrobert: the backend should verify that feeData is empty though, is that what you meant?

sunwrobert:
Closing this PR and replacing with https://github.com/ProjectOpenSea/opensea-js/pull/303

PR #284, opened 2021-12-14
chore: setup code-quality

Meemaw:
**Changes:**
- Add `prettier` && `prettier-package-json`
- Remove `lib` folder (built sources) from git
- Add basic `code_quality` CI:
  - Only checking Typescript types for now


sunwrobert:
qq: if we remove `lib`, then if we were to install, let's say like a local branch of opensea-js, it wouldn't work out of the box, right?

sunwrobert (approved):

PR #362, opened 2022-02-04
Support Wyvern 2.3 -- Bulk cancellations, EIP-712

sunwrobert:


package.json
```diff
@@ -39,8 +39,8 @@
     "lodash": "^4.17.21",
     "query-string": "^6.11.1",
     "web3": "0.20.7",
-    "wyvern-js": "git+https://github.com/ProjectOpenSea/wyvern-js.git#v3.3.4",
-    "wyvern-schemas": "git+https://github.com/ProjectOpenSea/wyvern-schemas.git#v0.14.0"
+    "wyvern-js": "git+https://github.com/ProjectOpenSea/wyvern-js.git#adb0b5ceb8616ce1414096a27c0a8a953120c0aa",
```
alexanderatallah, line 42: can we use git tags and versions for these? more readable and reviewable than commit hashes

package.json
```diff
@@ -39,8 +39,8 @@
     "lodash": "^4.17.21",
     "query-string": "^6.11.1",
```
alexanderatallah, line 40: can we make a version update for this package, to 2.0.0?

package.json
sunwrobert: I'm waiting for those PRs on the other repos to be approved and merged before doing that :P

package.json
sunwrobert: yup, will do

src/api.ts
```diff
@@ -121,6 +121,19 @@ export class OpenSeaAPI {
...
 
+  /**
+   * Get which version of Wyvern exchange to use to create orders
+   * Simply return null in case API doesn't give us a good response
+   */
+  public async getOrderCreateWyvernExchangeAddress(): Promise<string | null> {
+    try {
+      const result = await this.get(`${ORDERBOOK_PATH}/exchange/`);
+      return result as string;
+    } catch (error) {
```
alexanderatallah, line 132: can we log this error appropriately?

src/api.ts
```diff
@@ -121,6 +121,19 @@ export class OpenSeaAPI {
...
+      return null;
```
alexanderatallah, line 133: can we log the error appropriately?

src/seaport.ts
```diff
@@ -2543,8 +2570,12 @@ export class OpenSeaPort {
         useTxnOriginStaticCall: false,
       });
 
+    const exchange = await this._wyvernExchangeForOrderCreationPromise;
+
     return {
-      exchange: WyvernProtocol.getExchangeContractAddress(this._networkName),
+      exchange:
+        exchange ||
+        WyvernProtocol.getExchangeContractAddress(this._networkName),
```
alexanderatallah, line 2578: this looks like it's going to default to 2.3 if our API doesn't reply with any exchange address. can we have it default to the address passed in wyvernConfig? (could default to this if that isn't set either)

sunwrobert: sg

src/utils/schema.ts
```diff
@@ -121,8 +130,16 @@ export const encodeAtomicizedBuy: AtomicizedBuyEncoder = (
...
-export const encodeBuy: Encoder = (schema, asset, address) => {
-  const transfer = schema.functions.transfer(asset);
+export const encodeBuy: Encoder = (
+  schema,
+  asset,
+  address,
+  { validatorAddress } = {}
+) => {
+  const transfer =
+    validatorAddress && schema.functions.checkAndTransfer
```
alexanderatallah, line 140: this needs an english auction check

src/seaport.ts
```diff
@@ -3083,11 +3138,24 @@ export class OpenSeaPort {
...
       if ("asset" in order.metadata) {
         const schema = this._getSchema(order.metadata.schema);
         return order.side == OrderSide.Buy
-          ? encodeSell(schema, order.metadata.asset, recipientAddress)
-          : encodeBuy(schema, order.metadata.asset, recipientAddress);
+          ? encodeSell(
+              schema,
+              order.metadata.asset,
+              recipientAddress,
+              shouldValidate ? { validatorAddress: order.target } : undefined
```
alexanderatallah, line 3151: it seems like we have two ways of stopping the use of merkle validator: not passing in a config, like this, and passing in `isEnglishAuction`. should we just consolidate around the former?

alexanderatallah (approved):

src/constants.ts
```diff
@@ -63,3 +63,45 @@ export const MAINNET_PROVIDER_URL = `${API_BASE_MAINNET}/${RPC_URL_PATH}`;
...
 export const API_PATH = `/api/v${ORDERBOOK_VERSION}`;
+
+export const EIP_712_ORDER_TYPES = {
+  EIP712Domain: [
+    { name: "name", type: "string" },
+    { name: "version", type: "string" },
+    { name: "chainId", type: "uint256" },
+    { name: "verifyingContract", type: "address" },
+  ],
+  Order: [
```
mrbaguvix, line 74: Just curious, is there a way we can use `enum` type in any of these?

src/seaport.ts
```diff
@@ -2526,7 +2567,8 @@ export class OpenSeaPort {
     const { target, calldata, replacementPattern } = encodeBuy(
       schema,
       wyAsset,
-      accountAddress
+      accountAddress,
+      sellOrder ? undefined : merkleValidatorByNetwork[this._networkName]
```
alexanderatallah, line 2571: doesn't this need to be `sellOrder.isWaitingForCounterOrder`?

sunwrobert: ah, yeah i removed it to test something but can add it back

sunwrobert:
> Question: Does the public `fulfillOrder` method require changes to `_atomicMatch` to use EIP-712 and the `nonce`?

Nope, should be same

alexanderatallah (approved):

Issue #378, opened 2022-02-14
OpenSea-js 2.0.0 : createSellOrder() on Rinkeby leads to a 500 Error without API-Key


SMLaursen:
Hi,

When trying to call `createSellOrder()` I get an `500`. I consistently see that the `getOrderCreateWyvernExchangeAddress()` call fails due to throttling, which makes it use a default of `0xdd54d660178b28f6033a953b0e55073cfa7e3744` (Specified in `utils/deployed`) instead of the actual `0x5206e78b21ce315ce284fb24cf05e0585a93b1d9`.

Hardcoding `getOrderCreateWyvernExchangeAddress()` to always return `0x5206e78b21ce315ce284fb24cf05e0585a93b1d9` (without issuing any calls) makes everything work.

I think simple calls, such as these should be allowed through the throttling, possibly by adding retries/wait - and maybe the default of `0xdd54d660178b28f6033a953b0e55073cfa7e3744` also needs to be changed?

alexanderatallah:
Is this still a problem? we switched over to the new exchange today and there shouldn't be throttling on the exchange address (if there is, that's a bug). cc @sunwrobert

sunela:
![image](https://user-images.githubusercontent.com/88385740/154838244-6a0fd47f-d4b1-49ae-9424-f4d39dd6b37d.png)

I'm still getting throttled on the `https://testnets-api.opensea.io/wyvern/v1/exchange/?` request (Here shown with https://github.com/ProjectOpenSea/opensea-js/pull/382, such that it eventually succeeds)

sunela:
But now at least the default `0xdd54d660178b28f6033a953b0e55073cfa7e3744` is the same as what is returned by `https://testnets-api.opensea.io/wyvern/v1/exchange/?`

SMLaursen:
![image](https://user-images.githubusercontent.com/7354598/154838490-7579ca61-fbe1-4b33-8426-0c8e2e27ad40.png)

I'm still getting throttled on the https://testnets-api.opensea.io/wyvern/v1/exchange/? request (Here shown with https://github.com/ProjectOpenSea/opensea-js/pull/382, such that it eventually succeeds)

But now at least the default `0xdd54d660178b28f6033a953b0e55073cfa7e3744 `is the same as what is returned by `https://testnets-api.opensea.io/wyvern/v1/exchange/?`

alexanderatallah:
[...]

PR #382, opened 2022-02-16
Added retries to getOrderCreateWyvernExchangeAddress()

SMLaursen:
Fixes issue #378

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -126,15 +126,21 @@ export class OpenSeaAPI {
    * Get which version of Wyvern exchange to use to create orders
    * Simply return null in case API doesn't give us a good response
    */
-  public async getOrderCreateWyvernExchangeAddress(): Promise<string | null> {
+  public async getOrderCreateWyvernExchangeAddress(
+    retries = 2
+  ): Promise<string | null> {
     try {
-      const result = await this.get(`${ORDERBOOK_PATH}/exchange/`);
-      return result as string;
+      return await this.get<string>(`${ORDERBOOK_PATH}/exchange/`);
     } catch (error) {
-      this.logger(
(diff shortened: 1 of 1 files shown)


alexanderatallah:
This should be fully cached and non-limited - can you see if you're still being throttled?

src/api.ts
```diff
@@ -126,15 +126,22 @@ export class OpenSeaAPI {
    * Get which version of Wyvern exchange to use to create orders
    * Simply return null in case API doesn't give us a good response
    */
```
mrbaguvix, line 128: I believe we should update this doc since we added new logic in `try/catch` block.

src/api.ts
```diff
@@ -126,15 +126,22 @@ export class OpenSeaAPI {
...
-      );
-      return null;
+      if (retries <= 0) {
+        this.logger(
+          "Couldn't retrieve Wyvern exchange address for order creation"
+        );
+        return null;
+      } else {
+        await delay(3000);
+        return this.getOrderCreateWyvernExchangeAddress(retries - 1);
```
mrbaguvix, line 143: can you add a comment about this logic?

mrbaguvix (changes requested):

src/api.ts
SMLaursen: Done

src/api.ts
SMLaursen: Done

SMLaursen:
> This should be fully cached and non-limited - can you see if you're still being throttled?

I'm still being throttled, see my latest comment in https://github.com/ProjectOpenSea/opensea-js/issues/378

mrbaguvix (approved):

ryanio:
we no longer support wyvern, thanks

PR #388, opened 2022-02-18
Fix sign typed data to be compatible for more providers

sunwrobert:
Sign typed data v2,3,4 are not supported by a number of providers. Changing to just use regular signTypedData. Also removing JSON.stringify as some wallet providers sanitize the message

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "opensea-js",
-  "version": "2.0.0",
+  "version": "2.0.1-beta.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "opensea-js",
-  "version": "2.0.0",
+  "version": "2.0.1-beta.0",
```
sunwrobert, line 3: Updating this to be non beta version when this is approved

PR #434, opened 2022-03-17
Fix getAssets method to work with new API Cursor implementation

wcjohnson11:
The OpenSea API recently introduced [Cursor Pagination](https://docs.opensea.io/changelog/cursor-pagination), deprecating `offset` as a query param for the `assets` endpoint.

This made it so that the Opensea-js SDK was unable to query a dataset that would require using the `offset` param.

This PR removes the `offset` argument from the `getAssets` method, adds `next` and `previous` to the return and adds cursor as a prop in the `OpenSeaAssetQuery` type.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -232,7 +232,7 @@ export class OpenSeaAPI {
   public async getAssets(
     query: OpenSeaAssetQuery = {}
   ): Promise<{ assets: OpenSeaAsset[]; estimatedCount: number; next: string | undefined; previous: string | undefined }> {
-    const json = await this.get<{ estimated_count: number; assets: unknown[] }>(
+    const json = await this.get<{ estimated_count: number; assets: unknown[]; next: string | undefined; previous: string | undefined; }>(
       `${API_PATH}/assets/`,
       {
         limit: this.pageSize,


src/api.ts
```diff
@@ -228,24 +228,22 @@ export class OpenSeaAPI {
...
    * @param query Query to use for getting orders. A subset of parameters on the `OpenSeaAssetJSON` type is supported
-   * @param page Page number, defaults to 1. Can be overridden by
-   * `limit` and `offset` attributes from OpenSeaAssetQuery
    */
   public async getAssets(
-    query: OpenSeaAssetQuery = {},
-    page = 1
-  ): Promise<{ assets: OpenSeaAsset[]; estimatedCount: number }> {
+    query: OpenSeaAssetQuery = {}
+  ): Promise<{ assets: OpenSeaAsset[]; estimatedCount: number; next: string | undefined; previous: string | undefined }> {
```
michael-cohen-eth, line 234: Could we change the `page` param to instead be a `cursor` param that takes in the cursor token from previous runs? That way users of `getAssets` will be able to paginate properly. Unless the `query` param already takes care of this?

wcjohnson11: I suppose we could do that, but we'd need to add the cursor to the query object inside the function since that's the structure the API expects it in. I figured it would make sense to include all of the query params sent to the server in the query object passed to the `getAssets` method

You're a maintainer though so happy to refactor if that's what you'd prefer!

michael-cohen-eth: This makes sense to me, was just clarifying. I think that should be fine to leave as-is for now. If this becomes problematic we can adjust as needed. 

@sunwrobert thoughts on the discussion above? Incidentally we are definitely overdue for testing here.

sunwrobert: Makes sense to keep the current changes. and yes we are overdue for testing 😄

sunwrobert (approved):

src/api.ts
michael-cohen-eth: cool, approved as well

michael-cohen-eth (approved):

sunwrobert:
@wcjohnson11 can you fix lint issues?

sunwrobert:
Merged this into master, closing now.

PR #467, opened 2022-04-15
fix: remove balance check on buy order

Meemaw:


sunwrobert (approved):
There's a JSDoc in `createBundleBuyOrder` and `createBuyOrder` that says that it will throw the error, can we also remove that?

PR #488, opened 2022-04-23
Don't bundle in tests folder in lib

sunwrobert:
Decreases the unpacked size from 1.2 MB -> 850KB


Meemaw (commented):
My concern is that since we publish locally, one could still do a mistake and include it. Wdyt about adding it to npm ignore?

ts-config-build.json
```diff
@@ -0,0 +1,4 @@
+{
```
Meemaw, line 1: **nit:** lets name `tsconfig.build.json`.

Meemaw (approved):

package.json
```diff
@@ -18,7 +18,7 @@
   ],
   "scripts": {
     "abi-type-gen": "yarn run typechain --target=web3-v1 src/abi/*.json --out-dir=src/typechain/contracts",
-    "build": "yarn abi-type-gen && tsc",
+    "build": "yarn abi-type-gen && tsc -p ts-config.build.json",
```
Meemaw, line 21:
```suggestion
    "build": "yarn abi-type-gen && tsc -p tsconfig.build.json",
```

sunwrobert: oh woops

Meemaw (approved):

PR #505, opened 2022-04-29
fixes wrong getMaxOrderExpirationTimestamp calculation to use months …

agudpp:
…instead of days


agudpp:
@sunwrobert let me know if I should I do something else (related to the errors happening). I see develop is also failing, so probably related.

sunwrobert:
I'll get this pushed out over the weekend

JoshuaSchmidt-OpenSea:
This is fixed in master now.

PR #522, opened 2022-05-17
josh/remove-web3-dep

jtaaa:


carlosbalsas:
This will fix the warnings we are receiving like: `Compiled with problems:X

WARNING in ./node_modules/ethers/dist/ethers.min.js 1:484-485

Critical dependency: require function is used in a way in which dependencies cannot be statically extracted` right?

ethanleifer:
Hey! Is this pr still being worked on? I am getting an error when executing a buy order using opensea.js that is coming from web3.js and wyvern.js

Patientdame (changes requested):
_cancel all requests_

PR #536, opened 2022-05-26
Seaport protocol support

jtaaa:
**API Methods**
Updates API methods to use new endpoints that return new `OrderV2` format that support multiple protocols: `getOrder`, `getOrders`, `postOrder`

**SDK Methods**
Updated methods for cancelling, fulfilling, approving orders and creating orders to handle seaport protocol orders: `cancelOrder`, `fulfillOrder`, `approveOrder`, `isOrderFulfillable`, `createBuyOrder`, `createSellOrder`

Due to API support the following features are no longer supported with this release:
- `createBundleBuyOrder` and `createBundleSellorder`
- creating english auctions sell orders, i.e. `createSellOrder` options: `waitForHighestBid` and `englishAuctionReservePrice`

**Other**
- Adds `ethers` and `seaport-js` for interacting with seaport protocol data
- Adds `ajv` for order schema validation - currently only used in tests
- Adds `chai-as-promised` for testing promise rejections


src/seaport.ts
```diff
@@ -136,12 +149,18 @@ import {
...
   validateAndFormatWalletAddress,
   getMaxOrderExpirationTimestamp,
+  hasErrorCode,
+  getAssetItemType,
 } from "./utils/utils";
 
 export class OpenSeaPort {
   // Web3 instance to use
   public web3: Web3;
   public web3ReadOnly: Web3;
```
janclarin, line 159: Do we plan to remove Web3 later?

src/seaport.ts
```diff
@@ -649,6 +679,104 @@ export class OpenSeaPort {
...
+              collectionBuyerFeeBasisPoints,
+              asset.collection.payoutAddress
+            )
+          : undefined,
+    };
+  }
+
+  private getAssetItems(
+    assets: Asset[],
+    quantities: number[] = [],
```
janclarin, line 769: Should we pass these quantities in as bignumber? number isn't reliable for non-int quantities.

Applies everywhere we're passing around amounts

src/seaport.ts
```diff
@@ -731,6 +859,80 @@ export class OpenSeaPort {
...
+    accountAddress,
+    startAmount,
+    quantity = 1,
+    expirationTime,
+    paymentTokenAddress = WETH_ADDRESS,
+  }: {
+    asset: Asset;
+    accountAddress: string;
+    startAmount: number;
+    quantity?: number;
```
janclarin, line 884: Same - can we use bignumber?

src/seaport.ts
```diff
@@ -1149,6 +1436,48 @@ export class OpenSeaPort {
...
+   * @returns Transaction hash for fulfilling the order
+   */
+  public async fulfillOrder({
+    order,
+    accountAddress,
+  }: {
+    order: OrderV2;
+    accountAddress: string;
+    // TODO: Implement recipientAddress
+    recipientAddress?: string;
```
janclarin, line 1456: If it's not supported, maybe we should comment out the input param for now? Otherwise it could be used by accident if they don't read the comments.

src/seaport.ts
```diff
@@ -1206,6 +1548,43 @@ export class OpenSeaPort {
   public async cancelOrder({
     order,
     accountAddress,
+  }: {
+    order: OrderV2;
+    accountAddress: string;
+  }) {
+    // this._dispatch(EventType.CancelOrder, { order, accountAddress });
```
janclarin, line 1555: Is this supposed to be commented out?

src/seaport.ts
```diff
@@ -3430,7 +3859,38 @@ export class OpenSeaPort {
    * @param order Order to approve
    * @returns Transaction hash of the approval transaction
    */
-  public async approveOrder(order: UnsignedOrder) {
+  public async approveOrder(order: OrderV2) {
+    // this._dispatch(EventType.ApproveOrder, { order, accountAddress });
```
janclarin, line 3863: Same, commented out?

janclarin (commented):
Overall looks great!

src/seaport.ts
jtaaa: Yes we do!

src/seaport.ts
jtaaa: good point. I want the sdk consumer to be able to use a simple number when calling. I can make it take `BigNumberInput` and immediately convert to bignumber for internal use

src/seaport.ts
jtaaa: good catch. it is not

src/constants.ts
```diff
@@ -106,3 +106,15 @@ export const MERKLE_VALIDATOR_MAINNET =
...
+  "0x939c8d89ebc11fa45e576215e2353673ad0ba18a71286b93e3954e004a000000";
+export const OPENSEA_CROSS_CHAIN_CONDUIT =
+  "0x920000c70200b4006c20003a0e8a0030580800f1";
+
+export const CONDUIT_KEYS_TO_CONDUIT = {
+  [OPENSEA_CROSS_CHAIN_CONDUIT_KEY]: OPENSEA_CROSS_CHAIN_CONDUIT,
+};
+
+export const OPENSEA_ZONE = "0x9B814233894Cd227f561B78Cc65891AA55C62Ad2";
+export const WETH_ADDRESS = "0xc778417E063141139Fce010982780140Aa0cD5Ab";
```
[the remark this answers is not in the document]
jtaaa: Good catch they are not the same 🙏

kophyo1234 (approved):

sunwrobert (approved):
well done

wong2:
As an application developer, how do we handle both old oders and new orders?

PR #548, opened 2022-06-08
Updated README - Removal of old Affiliate info

itsthedoom:
Removal of Affiliate compensation reference in SDK.


jtaaa (approved):

koin-marian:
How do we set our referral wallet address with the new SDK? Has the Affiliate program been completely shut down?

PR #552, opened 2022-06-10
Remove search field from Asset & Bundle API lookups

michael-cohen-eth:
We don't support search as a query param in the Asset & Bundle endpoints anymore


jtaaa (approved):
Thanks @michael-cohen-eth can you check if we need to update docs as well?

michael-cohen-eth:
Good call, checked the README and didn't see anything, should be safe

Patientdame (commented):
please i would like to cancel the update and lookups

PR #553, opened 2022-06-10
Adds Optional Parameter to `fulfillOrder` Used for Percentage Based Gas Adjustments

QuitCrypto:
This is a QOL improvement for situations where you may want to use fast gas to buy a listing quickly. `gasMultiplier` is an optional parameter, so should not affect any existing implementations.

### Usage
Calling `fulfillOrder` with a gas multiplier is simple:
`await seaport.fulfillOrder({ order, accountAddress, gasMultiplier: 3 })`

### Additional considerations
It may be prudent to add a cap with `txnData.gasPrice = Math.min(txnData.gasPrice, maxGasPrice)` some something similar to avoid paying crazy amounts when gas is higher than expected. Could also consider using an additive rather than multiplicative modifier. 

### Late edit
After toying with this for a while, a better solution is to just pas a flat value in for gas. I do believe that maybe there could be some middle ground to allow for dynamic adjustment based on market conditions, but a flat gas value is what I ended up using the most when testing this locally, and is simpler for users to understand.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -1156,6 +1156,7 @@ export class OpenSeaPort {
    * @param accountAddress The taker's wallet address
    * @param recipientAddress The optional address to receive the order's item(s) or curriencies. If not specified, defaults to accountAddress.
    * @param referrerAddress The optional address that referred the order
+   * @param gasMultiplier An option for submitting transaction with higher or lower than estimated gwei
    * @returns Transaction hash for fulfilling the order
    */
   public async fulfillOrder({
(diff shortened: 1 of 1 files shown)


asafd11:
is there a way to add extra to the gas right now ?  with the new version 4.0.3
i tryd 
```
// Extra gwei to add to the mean gas price when making transactions
   this.gasPriceAddition = new bignumber_js_1.BigNumber(30);  // was 3 be default 
  // Multiply gas estimate by this factor when making transactions
  this.gasIncreaseFactor = constants_2.DEFAULT_GAS_INCREASE_FACTOR;  // i set it to 2 default was 1
```
not working not effecting the gas

QuitCrypto:
> is there a way to add extra to the gas right now ? with the new version 4.0.3 i tryd
> 
> ```
> // Extra gwei to add to the mean gas price when making transactions
>    this.gasPriceAddition = new bignumber_js_1.BigNumber(30);  // was 3 be default 
>   // Multiply gas estimate by this factor when making transactions
>   this.gasIncreaseFactor = constants_2.DEFAULT_GAS_INCREASE_FACTOR;  // i set it to 2 default was 1
> ```
> 
> not working not effecting the gas

I don't believe there is at the moment

PR #561, opened 2022-06-14
chore: Update default opensea zone

jtaaa:


src/constants.ts
```diff
@@ -117,7 +117,7 @@ export const CONDUIT_KEYS_TO_CONDUIT = {
   [CROSS_CHAIN_DEFAULT_CONDUIT_KEY]: CROSS_CHAIN_DEFAULT_CONDUIT,
 };
 
-export const DEFAULT_ZONE = "0x9b814233894cd227f561b78cc65891aa55c62ad2";
+export const DEFAULT_ZONE = "0x004C00500000aD104D7DBd00e3ae0A5C00560C00";
```
sunwrobert, line 120: This zone is only applicable for Ethereum.

src/constants.ts
```diff
@@ -117,7 +117,7 @@ export const CONDUIT_KEYS_TO_CONDUIT = {
   [CROSS_CHAIN_DEFAULT_CONDUIT_KEY]: CROSS_CHAIN_DEFAULT_CONDUIT,
 };
 
-export const DEFAULT_ZONE = "0x9b814233894cd227f561b78cc65891aa55c62ad2";
+export const DEFAULT_ZONE = "0x004c00500000ad104d7dbd00e3ae0a5c00560c00";
```
sunwrobert, line 120: This zone is only applicable for Ethereum, and it changes based on if it's an english auction or not

English auction uses old zone 0x9b814233894cd227f561b78cc65891aa55c62ad2, all other orders use new zone only for ethereum.

jtaaa: We don't support posting english auctions through the UI right now. Does that mean rinkeby should use old zone?

sunwrobert: Rinkeby should just use the old zone for now.

jtaaa: yeah I mean we don't support posting english auctions through the new orders endpoints

sunwrobert: Hm, okay well then yeah let's just change it based on network then

sunwrobert (approved):

PR #599, opened 2022-06-21
fix: Restrict by zone and allow partial fills by default

jtaaa:


src/sdk.ts
```diff
@@ -882,13 +883,15 @@ export class OpenSeaSDK {
     quantity = 1,
     expirationTime,
     paymentTokenAddress,
+    allowPartialFills = true,
```
sunwrobert, line 886: I wonder if we should even be customizing this. Currently all ERC-1155 orders are partially fulfillable, and we don't have any logic to deal with orders that are set to be non-partially filled.

jtaaa: ok in that case I wont make it customizable and will always pass true. cc @cterech for awareness

PR #602, opened 2022-06-22
chore: Remove /api prefix from API path

jtaaa:


sunwrobert (approved):

asghaier76:
@jtaaa now the request is without the api prefix but it is returning CORS error

Issue #624, opened 2022-06-27
Complete checkout in frontend returns execution reverted when order was created with sdk 


Lebski:
When I create an order with `createSellOrder` I can see the order on opensea.io, but I'm not able to purchase it. 

The returned in the console is: 
```Javascript
 "originalError": {
    "code": 3,
    "data": "0xfb5014fc2377ca19c4abe0fc1118f9526fd800bf94c1a012c593f07abf9bc0941c8fafa8",
    "message": "execution reverted"
  }
 ```

and originates in https://rinkeby.etherscan.io/address/0x00000000006c3852cbef3e08e8df289169ede581

When I decode the error I get: 
```javascript
{
  "method": "InvalidRestrictedOrder",
  "types": [
    "bytes32"
  ],
  "inputs": [
    "0xa5addc997d7ca41bcd0395e93e279e8436fabe2a21501e7145f5a0ea7ab2cfa5"
  ],
  "names": [
    "orderHash"
  ]
}
```

Code which was used to create the order: 

```javascript
const factorySellOrder = await seaport.createSellOrder({
            asset:
            {
                tokenId: tokenId,
                tokenAddress: tokenAddress,
                schemaName: WyvernSchemaName.ERC721,
                version: TokenStandardVersion.ERC721v3
            },
            // Also not work without schemaName
            accountAddress: sellerAddress,
            startAmount: price,
            listingTime: startUnix,
            expirationTime: endUnix,
        });
```
SDK version: 4.0.3

![image](https://user-images.githubusercontent.com/25309247/176033649-02745a21-e37f-41fe-8883-aa0540ee4133.png)

kicker1337:
same issue execution reverted { "originalError": { "code": 3, "data": "0xfb5014fca0bac144798db5c052f351b053c9c310d28a01a3b2f761165f71288f8b8e68bf", "message": "execution reverted" } }

gogogocheng:
same issue(but with createBuyOrder)
execution reverted { "originalError": { "code": 3, "data": "0xfb5014fc07b7be4ddbfee3798d5ac735eac921cee71272946f10b56011221bfd7e3dfb72", "message": "execution reverted" } }

Lebski:
Did you use an basic EREC721 contract? @gogogocheng and @kicker1337 ?

asghaier76:
[...]

PR #633, opened 2022-06-29
Update DEFAULT_ZONE_BY_NETWORK (rinkeby)

plondon:
While `0x9b814233894cd227f561b78cc65891aa55c62ad2` zone does not throw an error when trying to create an order, it does throw execution reverted when trying to fulfill that order. #624


jtaaa (approved):
🙏

PR #695, opened 2022-08-17
Update Fee Recipient to the Fee Collector

BCLeFevre:
This is updating the Fee Recipient to the Fee Collector that is at address `0x0000a26b00c1F0DF003000390027140000fAa719`.

Dev Comms will need to be given to inform users to update to the latest SDK before we stop accepting orders with the legacy fee recipient address.


sunwrobert (approved):

sunwrobert:
> This is updating the Fee Recipient to the Fee Collector that is at address `0x0000a26b00c1F0DF003000390027140000fAa719`.
> 
> Dev Comms will need to be given to inform users to update to the latest SDK before we stop accepting orders with the legacy fee recipient address.

Unfortunately this is already active, so we need to update now.

BCLeFevre:
> > This is updating the Fee Recipient to the Fee Collector that is at address `0x0000a26b00c1F0DF003000390027140000fAa719`.
> > Dev Comms will need to be given to inform users to update to the latest SDK before we stop accepting orders with the legacy fee recipient address.
> 
> Unfortunately this is already active, so we need to update now.

While this is active, we should still be accepting orders that use our old fee recipient as well.

robwilkes:
> > > This is updating the Fee Recipient to the Fee Collector that is at address `0x0000a26b00c1F0DF003000390027140000fAa719`.
> > > Dev Comms will need to be given to inform users to update to the latest SDK before we stop accepting orders with the legacy fee recipient address.
> > 
> > 
> > Unfortunately this is already active, so we need to update now.
> 
> While this is active, we should still be accepting orders that use our old fee recipient as well.

you were not, functionality broke for many people

PR #725, opened 2022-09-03
Update consideration items for multiple royalty collections

Racherin:
modify incorrect elements for considerationFeeItems array.


JoshuaSchmidt-OpenSea:
Function no longer exists

PR #728, opened 2022-09-06
a cleaner way to make sure `Fees` submap is the correct type

sunnyshapir:
No functionality change - just cleaner syntax :)

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -726,7 +726,7 @@ export class OpenSeaSDK {
     const getConsiderationItemsFromSellerFees = (
       fees: Fees
     ): ConsiderationInputItem[] => {
-      const sellerFees = fees?.sellerFees as Map<string, number>;
+      const sellerFees = fees?.sellerFees;
       return Object.keys(sellerFees).map((recipient) =>
         getConsiderationItem(sellerFees.get(recipient) || 0, recipient)
       );
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -726,7 +726,7 @@ export class OpenSeaSDK {
     const getConsiderationItemsFromSellerFees = (
       fees: Fees
     ): ConsiderationInputItem[] => {
-      const sellerFees = new Map(Object.entries(fees?.sellerFees));
+      const sellerFees = fees?.sellerFees as Map<string, number>;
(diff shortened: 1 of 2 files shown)


src/sdk.ts
```diff
@@ -726,7 +726,7 @@ export class OpenSeaSDK {
     const getConsiderationItemsFromSellerFees = (
       fees: Fees
     ): ConsiderationInputItem[] => {
-      const sellerFees = new Map(Object.entries(fees?.sellerFees));
+      const sellerFees = fees?.sellerFees as Map<string, number>;
```
cterech, line 729: did you test this?

sunnyshapir: yep covered in `fees.ts` from a prior pr

cterech (approved):

src/sdk.ts
royliu2, line 729: might be a dumb question but isn't `sellerFees` already a map according to the types? why do we need to coerce this here

royliu2 (approved):

src/sdk.ts
sunnyshapir: realizing now, i think alternative #2 in this pr description is required for that? https://github.com/ProjectOpenSea/opensea-js/pull/727 
this might be the better approach to avoid having to do the coercion anywhere else

sunwrobert: hmm, yeah I don't think we should be doing any manual casting for cleanliness

sunnyshapir: sounds good, i'll update the pr to just set it properly in `collectionFromJSON`

src/sdk.ts
```diff
@@ -726,7 +726,7 @@ export class OpenSeaSDK {
     const getConsiderationItemsFromSellerFees = (
       fees: Fees
     ): ConsiderationInputItem[] => {
-      const sellerFees = new Map(Object.entries(fees?.sellerFees));
+      const sellerFees = fees?.sellerFees;
```
sunwrobert, line 729: nit: unnecessary `?`

sunwrobert (approved):

0xIsshiki:
I think there might still be a bug in here. See https://github.com/ProjectOpenSea/opensea-js/pull/731 for my attempt to fix it.

PR #731, opened 2022-09-09
Iterate over sellerFees Map correctly

0xIsshiki:
`Object.keys` over a `Map` doesn't work; it will return an empty array.

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -727,9 +727,9 @@ export class OpenSeaSDK {
       fees: Fees
     ): ConsiderationInputItem[] => {
       const sellerFees = fees.sellerFees;
-      return Object.keys(sellerFees).map((recipient) =>
-        getConsiderationItem(sellerFees.get(recipient) || 0, recipient)
-      );
+      return Array.from(sellerFees.entries()).map(([recipient, _]) => {
+        return getConsiderationItem(sellerFees.get(recipient) || 0, recipient);
+      });
     };
 
     return {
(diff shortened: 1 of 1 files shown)


src/sdk.ts
```diff
@@ -727,9 +727,9 @@ export class OpenSeaSDK {
       fees: Fees
     ): ConsiderationInputItem[] => {
       const sellerFees = fees.sellerFees;
-      return Object.keys(sellerFees).map((recipient) =>
-        getConsiderationItem(sellerFees.get(recipient) || 0, recipient)
-      );
+      return Array.from(sellerFees.entries()).map(([recipient, _]) => {
+        return getConsiderationItem(sellerFees.get(recipient) || 0, recipient);
+      });
```
sunwrobert, line 732:
```suggestion
      return Array.from(sellerFees.entries()).map(([recipient, basisPoints]) => {
        return getConsiderationItem(basisPoints, recipient);
      });
```

sunwrobert (approved):
Thanks for this! cc @sunnyshapir 

LGTM after suggested changes

sunwrobert:
> Thanks for this! cc @sunnyshapir
> 
> LGTM after suggested changes

Can you lint this actually? Then we can merge

0xIsshiki:
> > Thanks for this! cc @sunnyshapir
> > LGTM after suggested changes
> 
> Can you lint this actually? Then we can merge

Done!

Issue #734, opened 2022-09-11
Error: invalid BigNumber value


jackwolt:
Started getting this error today, on openseaSDK.createSellOrder() method with Node 16.14.0

```
await openseaSDK.createSellOrder({
        asset: {
             tokenId: item.id,
        },
        startAmount: item.startAmount, // 1.5
        endAmount: item.endAmount,  // 1.35  
        ...

        /** Getting id, prices - from my "item" array  */
 });
```

My current package.json
```
"dependencies": {
    "yarn": "^1.22.19",
    "@0x/subproviders": "^7.0.0",
    "@truffle/hdwallet-provider": "^2.0.14",
    "express": "^4.18.1",
    "opensea-js": "^4.0.8",
    "web3": "^1.7.5"
  }
```

Full error information.
```
Error: invalid BigNumber value (argument="value", value=undefined, code=INVALID_ARGUMENT, version=bignumber/5.6.2)
    at Logger.makeError (/app/node_modules/@ethersproject/logger/lib/index.js:233:21)
    at Logger.throwError (/app/node_modules/@ethersproject/logger/lib/index.js:242:20)
    at Logger.throwArgumentError (/app/node_modules/@ethersproject/logger/lib/index.js:245:21)
    at Function.BigNumber.from (/app/node_modules/@ethersproject/bignumber/lib/bignumber.js:239:23)
    at toBN (/app/node_modules/@ethersproject/bignumber/lib/bignumber.js:292:25)
    at BigNumber.add (/app/node_modules/@ethersproject/bignumber/lib/bignumber.js:57:43)
    at /app/node_modules/@opensea/seaport-js/lib/utils/order.js:151:43
    at Array.reduce (<anonymous>)
    at totalItemsAmount (/app/node_modules/@opensea/seaport-js/lib/utils/order.js:147:10)
    at Seaport.<anonymous> (/app/node_modules/@opensea/seaport-js/lib/seaport.js:285:76)
    at step (/app/node_modules/@opensea/seaport-js/lib/seaport.js:44:23)
    at Object.next (/app/node_modules/@opensea/seaport-js/lib/seaport.js:25:53)
    at fulfilled (/app/node_modules/@opensea/seaport-js/lib/seaport.js:16:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
        reason: 'invalid BigNumber value',
        code: 'INVALID_ARGUMENT',
        argument: 'value',
        value: undefined
    }
```

[...]

PR #740, opened 2022-09-14
[bug fix] Spread over collection seller fees

0xIsshiki:
The considerationFeeItems array needs to be flat.

Fixes https://github.com/ProjectOpenSea/opensea-js/issues/734.

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -833,7 +833,7 @@ export class OpenSeaSDK {
       });
     const considerationFeeItems = [
       openseaSellerFee,
-      collectionSellerFees,
+      ...collectionSellerFees,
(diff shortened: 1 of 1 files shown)


mr-friend (changes requested):
suggest using `.flat()` instead, `collectionSellerFees` can be `undefined` at this stage.
>const considerationFeeItems = [
      openseaSellerFee,
      collectionSellerFees,
      collectionSellerFees,
    ].filter((item): item is ConsiderationInputItem => item !== undefined)`.flat()`;

sunwrobert:
Fixing this in this PR: #746 thanks for bringing this up though.

PR #752, opened 2022-09-21
fix: sell order consideration items

sunwrobert:
I did the fix for the buy order, but not the sell order 😓


sunwrobert:
Verified this manually. Should be able to create sell orders now.

PR #762, opened 2022-09-30
update fulfillment methods to accept domain

stephankmin:
- Add optional `domain` param to fulfillment methods to match functionality in `seaport-js` [PR](https://github.com/ProjectOpenSea/seaport-js/pull/117)
- Appends tags to fulfillment calldata


src/sdk.ts
```diff
@@ -947,9 +947,11 @@ export class OpenSeaSDK {
   private async fulfillPrivateOrder({
     order,
     accountAddress,
+    domain = "",
```
jtaaa, line 950: nit: don't think we need the default value

jtaaa (approved):

PR #774, opened 2022-10-14
chore: remove english auction documentation as its not supported yet

sunwrobert:


jtaaa (approved):

nl0243:
What would be the best way to set an English auction for several 1000s NFTs, since API does not support this functionality yet?

PR #779, opened 2022-10-25
add header to requests for tracking

sunnyshapir:
i'm not familiar with best practices when using custom tags so definitely open to suggestions, but just adding something to help gain some visibility into where certain requests are coming from


cterech (approved):
LGTM assuming we test it and the header gets through to DD

JoshuaSchmidt-OpenSea:
Added a new tag to requests for tracking

PR #844, opened 2023-02-28
Seaport v1.4 functionality

ryanio:
- Updates to seaport-js 1.0.9 with seaport v1.4 functionality
- Changes to create orders by default on seaport v1.4, unless v1.1 is passed into order.protocolAddress
- Bumps opensea-js version to 4.0.15


cterech (approved):

pravin120993:
`if (
      !(
        order.protocolAddress in
        [CROSS_CHAIN_SEAPORT_ADDRESS, CROSS_CHAIN_SEAPORT_V1_4_ADDRESS]
      )
    ) {
      throw new Error("Unsupported protocol");
    }`
    
   this is wrong `in` operator will check the index. this condition will be alway false. because of this i am getting `Unsupported protocol`

PR #850, opened 2023-03-02
Add sdk version as header in API requests

JoshuaSchmidt-OpenSea:
Sending the SDK version in our requests will make it quicker to diagnose issues which are only seen in the SDK and not the API.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -392,7 +392,7 @@ export class OpenSeaAPI {
       ...opts,
       headers: {
         ...(apiKey ? { "X-API-KEY": apiKey } : {}),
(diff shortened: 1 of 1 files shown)


src/api.ts
```diff
@@ -391,6 +392,7 @@ export class OpenSeaAPI {
       ...opts,
       headers: {
         ...(apiKey ? { "X-API-KEY": apiKey } : {}),
+        ...{ "sdk-version": sdkVersion },
```
montasaurus, line 395:
```suggestion
        ...{ "SDK-Version": sdkVersion },
```

Just to match the usual header format

montasaurus (approved):

PR #853, opened 2023-03-05
fix: protocal address condition fix

pravin120993:


pi9min:
@pravin120993 

Great PR! I have been waiting for it too.
One point, the comparison with includes is case sensitive, and a comparison of an EIP-55 style address with check sum and a lower-case address will return false, even if they are the same address.
In this case, it seems better to compare all addresses with `toLowerCase()`.

e.g. 
```
export const isSupportedProtocolAddress = (address: string): boolean => {
  return [CROSS_CHAIN_SEAPORT_ADDRESS, CROSS_CHAIN_SEAPORT_V1_4_ADDRESS].map((a) => a.toLowerCase()).includes(address.toLowerCase())
}

...
    if (!isSupportedProtocolAddress(order.protocolAddress)) {
...
```

Please check it out.
Thanks.

PR #857, opened 2023-03-06
Compare Protocols By Address

JoshuaSchmidt-OpenSea:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "opensea-js",
-  "version": "4.0.15",
+  "version": "4.0.16",
```
cterech, line 3: i think this version is already out - have you rebased?

cterech: https://www.npmjs.com/package/opensea-js

PR #868, opened 2023-03-08
collection offers

cterech:


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,37 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
```
JoshuaSchmidt-OpenSea, line 1: What happens if a test run fails?

JoshuaSchmidt-OpenSea (approved):

.github/workflows/npm-publish.yml
cterech: ill create some docs for it

Issue #870, opened 2023-03-09
FulfillOrder will break when signatures are stripped


SMLaursen:
The OrderWithCounter object (protocolData) passed to `SeaPort.fulfillOrder` doesn't seem to be updated with signatures from `this.api.generateFulfillmentData` - instead the `order.clientSignature` is set, which I don't believe is used for fulfilling

PR #871, opened 2023-03-09
fix fulfillment

cterech:
Issue is brought up here: https://github.com/ProjectOpenSea/opensea-js/issues/870

We use ```seaport.fulfillOrder``` to fulfill orders, and we pass the ```order.protocolData``` to that method. The protocolData needs to have a set signature, and I previously only update the orderV2 object

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -1039,14 +1039,16 @@ export class OpenSeaSDK {
       throw new Error("Unsupported protocol");
     }
 
-    if (!order.clientSignature && order.orderHash) {
+    if (order.orderHash) {
       const result = await this.api.generateFulfillmentData(
         accountAddress,
         order.orderHash,
         order.protocolAddress,
         order.side
       );
-      order.clientSignature = result.fulfillment_data.orders[0].signature;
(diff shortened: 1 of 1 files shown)


src/sdk.ts
```diff
@@ -1039,14 +1039,16 @@ export class OpenSeaSDK {
...
       const result = await this.api.generateFulfillmentData(
         accountAddress,
         order.orderHash,
         order.protocolAddress,
         order.side
       );
-      order.clientSignature = result.fulfillment_data.orders[0].signature;
+      const signature = result.fulfillment_data.orders[0].signature;
+      order.clientSignature = signature;
+      order.protocolData.signature = signature;
```
JoshuaSchmidt-OpenSea, line 1051: Can we pull this out to a separate function so it can be unit tested? A function which takes in accountAddress and order and outputs OrderWithCounter.

cterech: @JoshuaSchmidt-OpenSea writing an integration test for fulfillment is a bit nuanced - you need to create a listing with one wallet and fulfill it with another

JoshuaSchmidt-OpenSea (approved):

PR #872, opened 2023-03-09
Remove wyvern-schemas package

JoshuaSchmidt-OpenSea:
Removing this package fixes known security issues with the version of `axios` used by the wyvern-schemas package


package.json
```diff
@@ -105,5 +106,12 @@
     "**/*.{ts,tsx,js,jsx}": [
       "eslint --cache --fix"
     ]
+  },
```
JoshuaSchmidt-OpenSea, line 109: 👀

src/sdk.ts
```diff
@@ -608,7 +609,7 @@ export class OpenSeaSDK {
     amountInEth: number;
     accountAddress: string;
   }) {
-    const token = WyvernSchemas.tokens[this._networkName].canonicalWrappedEther;
+    const token = getCanonicalWrappedEther(this._networkName);
```
JoshuaSchmidt-OpenSea, line 612: 👀

src/utils/tokens/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  address: string;
+}
+
+export interface NetworkTokens {
+  canonicalWrappedEther: Token;
+  otherTokens: Token[];
+}
+
+// TODO: Add Tests
+export const getCanonicalWrappedEther = function (network: Network): Token {
```
JoshuaSchmidt-OpenSea, line 19: 👀

src/utils/schemas/main/CryptoKitties/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  FunctionInputKind,
+  FunctionOutputKind,
+  StateMutability,
+} from "wyvern-js/lib/types";
+import { Schema } from "../../schema";
+import { EventInputKind } from "../../types";
+
+type CryptoKittiesType = string;
+
+export const CryptoKittiesSchema: Schema<CryptoKittiesType> = {
```
cterech, line 13: do we actually need this?

src/utils/schemas/main/CryptoPunks/index.ts
```diff
@@ -0,0 +1,106 @@
...
+  FunctionOutputKind,
+  StateMutability,
+} from "wyvern-js/lib/types";
+import { Schema } from "../../schema";
+import { EventInputKind } from "../../types";
+
+type CryptoPunksType = string;
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export const CryptoPunksSchema: Schema<CryptoPunksType> = {
```
cterech, line 13: do we need this? what are these collection specific objects used for?

cterech (commented):
few higher level comments
1. can we get rid of the collection specific definitions? i.e. why do we need a CryptoKitties schema? 
2. is it a lot of additional work to remove all of the references to Rinkeby? 
3. (this might be a dumb question) - instead of copying and pasting all of these object definitions, is there a way we can define our own minimal set of schemas and use those?

src/utils/schemas/main/CryptoKitties/index.ts
JoshuaSchmidt-OpenSea: We would need this if someone creates an order for CyptoKitties

JoshuaSchmidt-OpenSea:
> few higher level comments
> 
> 1. can we get rid of the collection specific definitions? i.e. why do we need a CryptoKitties schema?
> 2. is it a lot of additional work to remove all of the references to Rinkeby?
> 3. (this might be a dumb question) - instead of copying and pasting all of these object definitions, is there a way we can define our own minimal set of schemas and use those?

I think we will be able to remove CryptoPunks, Rinkeby and possible a set of the schemas and tokens. However, I would like to not make any changes to the logic in this PR since it is already massive just by removing the package.

package.json
```diff
@@ -22,7 +22,7 @@
     "check-types": "tsc --noEmit --project tsconfig.json",
     "coverage-report": "nyc report --reporter=text-lcov | coveralls",
     "docs-build": "typedoc --out docs src/index.ts",
-    "eslint:check": "eslint . --max-warnings 0 --ext .js,.ts",
+    "eslint:check": "eslint . --fix --max-warnings 0 --ext .js,.ts",
```
montasaurus, line 25: Don't think we want this here--should fail instead of fixing. Could add another `eslint:fix` if you want

montasaurus (approved):

cterech (approved):

PR #891, opened 2023-03-29
Remove wyvern usage for validating signatures and converting to base …

JoshuaSchmidt-OpenSea:


cterech (approved):
lgtm - only thing that gives me heartburn is the lack of test coverage on wrapping / unwrapping ETH but should be fine

PR #935, opened 2023-04-28
Use Seaport v1.5 for Listings and Offers

JoshuaSchmidt-OpenSea:


src/orders/utils.ts
```diff
@@ -46,6 +47,7 @@ export const getPostCollectionOfferPayload = (
       collection: { slug: collectionSlug },
     },
     protocol_data,
+    protocol_address: SEAPORT_CONTRACT_VERSION_V1_5,
```
ryanio, line 50: instead of adding this here, wondering if you could define a `DEFAULT_SEAPORT_CONTRACT_VERSION = SEAPORT_CONTRACT_VERSION_V1_5` and use that everywhere, then in the future just have to change one line to update version

ryanio (approved):
overall lgtm! see my comment for a non blocking suggestion

cterech (approved):
lgtm - easiest way to test would probably be to use this https://github.com/ProjectOpenSea/opensea-js/blob/master/src/__integration_tests__/README.md

```
API_KEY="..." WALLET_ADDRESS="..." WALLET_PRIV_KEY="..." ALCHEMY_API_KEY="..." npm run integration_tests

```

montasaurus (approved):

PR #940, opened 2023-05-03
fix dependencies @opensea/seaport-js to ^1.2.0

an-ivannikov:
@opensea/seaport-js version  v1.1.0 is automatically installed.    But it does not contain the constant CROSS_CHAIN_SEAPORT_V1_5_ADDRESS - It throws an error 'Unsupported protocol' for protocolAddress = 0x00000000000000adc04c56bf30ac9d3c0aaf14dc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   },
   "types": "lib/index.d.ts",
   "dependencies": {
-    "@opensea/seaport-js": "^1.0.10",
+    "@opensea/seaport-js": "^1.2.0",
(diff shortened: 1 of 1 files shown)


ryanio:
thanks for reporting, we will look into it

ryanio:
@an-ivannikov i just checked and the package-lock was updated with 1.2.0 in the v5 release, since it covers within the carat ^ range, can you try running a fresh `npm i`? that should fix for you

PR #948, opened 2023-05-12
Define Type for OpenseaCollection

CryptoUncleSam:
I am using the module with `OpenSeaAPI` on ts

When I use `.getCollection( ) ` and go to` .stats.somethingElse`. This "something else" is not define because the type is just object so I add the type that come from response.


ryanio:
could you define it without adding a lib folder, so just the `src/types.ts` change?

CryptoUncleSam:
opened a new one just types

PR #949, opened 2023-05-12
Add integration test for createBuyOrder

JoshuaSchmidt-OpenSea:


src/__integration_tests__/postOrder.ts
```diff
@@ -35,15 +35,29 @@ const sdk = new OpenSeaSDK(
 );
 
 suite("SDK: order posting", () => {
+  test("Post Buy Order", async () => {
+    const postBuyOrder = {
+      accountAddress: WALLET_ADDRESS ? WALLET_ADDRESS : "",
+      startAmount: "0.004",
+      asset: {
+        tokenAddress: "0x1a92f7381b9f03921564a437210bb9396471050c",
```
cterech, line 43: i grappled with this when i implemented - should we make the tester pass these in?

cterech: maybe have them pass it in optionally and default if they dont

src/__integration_tests__/postOrder.ts
```diff
@@ -35,15 +35,29 @@ const sdk = new OpenSeaSDK(
...
+        tokenId: "2288",
+      },
+    };
+
+    await sdk.createBuyOrder(postBuyOrder);
```
cterech, line 48: can we check that the order actually got posted? either in the response or do a 'get'

cterech: currently just relying on it throwing an exception or not

src/__integration_tests__/postOrder.ts
JoshuaSchmidt-OpenSea: We probably should pass in the start amount because the tester will need to have that much WETH but the token address shouldn't matter. I just chose the rarest cool-cat

cterech: yea fair enough

cterech (approved):

PR #955, opened 2023-05-16
Add X-Requested-With header to API calls

JoshuaSchmidt-OpenSea:
Adding header so we can determine how many calls are made via the SDK versus the API

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -469,6 +469,7 @@ export class OpenSeaAPI {
       ...opts,
       headers: {
         ...(apiKey ? { "X-API-KEY": apiKey } : {}),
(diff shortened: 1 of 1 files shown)


src/api.ts
```diff
@@ -469,6 +469,7 @@ export class OpenSeaAPI {
       ...opts,
       headers: {
         ...(apiKey ? { "X-API-KEY": apiKey } : {}),
+        ...{ "X-Requested-With": "Opensea-JS" },
```
Meemaw, line 472: Why are we creating a nested object and spreading it? Can be a plain key: value in the `headers` object.

PR #980, opened 2023-05-22
Remove dependency on axios

JoshuaSchmidt-OpenSea:
Remove package which was only used by CryptoKitties.

### DIFF
--- a/src/utils/schemas/main/CryptoKitties/index.ts
+++ b/src/utils/schemas/main/CryptoKitties/index.ts
@@ -1,4 +1,3 @@
-import axios from "axios";
 import { Schema } from "../../schema";
 import {
   AbiType,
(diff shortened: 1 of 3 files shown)


src/utils/schemas/main/CryptoKitties/index.ts
```diff
@@ -22,18 +21,18 @@ export const CryptoKittiesSchema: Schema<CryptoKittiesType> = {
...
-        if (
-          err.response &&
-          (err.response.status === 404 || err.response.status === 400)
-        ) {
-          return null;
-        } else {
-          throw err;
-        }
-      });
+    const response = await fetch(
```
cterech, line 24: do we even need this custom logic for cyrpto kitties?

cterech (approved):

src/utils/schemas/main/CryptoKitties/index.ts
JoshuaSchmidt-OpenSea: Unfortunately, I think we do because:
1. CryptoKitties are traded on OpenSea.
2. We have a `getBalance` function which calls the on-chain contracts
3. The `getBalance` function needs to get the schema
4. CryptoKitties was a contract released before modern erc721 formats

PR #988, opened 2023-05-24
Update stale.yml

ryanio:
temporarily bump to 1000 runs to get through all issues


JoshuaSchmidt-OpenSea (approved):
Isn't there only 261 issues? Shouldn't this finish by EOW if it runs once per day as-is?

ryanio:
it only gets through 100 api reqs so it keeps doing the same issues over

PR #1015, opened 2023-06-01
make postinstall script nonblocking on error

ryanio:


JoshuaSchmidt-OpenSea (approved):
We should also bump the patch version `6.0.1` so people can consume it

PR #1016, opened 2023-06-01
Remove URLSearchParams import

### LINKED ISSUE #1011: url_1.URLSearchParams is not a constructor
Looks like a bad import in api.ts in version 5.0.3

```
import { URLSearchParams } from "url";
```

JoshuaSchmidt-OpenSea:
Fixes #1011 

The import causes issues in ReactJS. The fix is to remove the import. See https://stackoverflow.com/a/58189753

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "opensea-js",
-  "version": "6.0.1",
+  "version": "6.0.2",
   "description": "JavaScript SDK for the OpenSea marketplace helps developers build new experiences using NFTs and our marketplace data!",
   "license": "MIT",
   "author": "OpenSea Developers",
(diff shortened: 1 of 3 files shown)


ryanio (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "opensea-js",
-  "version": "6.0.1",
+  "version": "6.0.2",
```
ryanio, line 3: ah didn't realize we already did 6.0.1, can you make a smol gh release for it? Ideally you can just use the npm publish workflow so you make the release to publish on npm automatically

denizsult:
Can you release this pr?

PR #1031, opened 2023-06-08
Fixes underflow when try to use fractional ETH as number in orders

### LINKED ISSUE #1022: Number underflow when try to use fractional ETH in order
```
		const listing = await openseaSDK.createSellOrder({
			asset: {
				tokenId,
				tokenAddress,
			},
			accountAddress,
			startAmount: 0.1,
			expirationTime
		})
```
Causes error

`
underflow [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT-underflow ] (fault="underflow", operation="BigNumber.from", value=0.1, code=NUMERIC_FAULT, version=bignumber/5.7.0)
`

JoshuaSchmidt-OpenSea:
## Motivation

Fixes #1022 

## Solution

BigNumberish was being parsed as FixedNumber and cast to string. Instead casted BigNumberish to string directly.

## Testing

Changed integration tests to use numbers instead of strings.

### DIFF
--- a/test/integration/postOrder.spec.ts
+++ b/test/integration/postOrder.spec.ts
@@ -15,7 +15,7 @@ suite("SDK: order posting", () => {
   test("Post Buy Order", async () => {
     const buyOrder = {
       accountAddress: walletAddress,
-      startAmount: OFFER_AMOUNT,
+      startAmount: +OFFER_AMOUNT,
       asset: {
         tokenAddress: "0x1a92f7381b9f03921564a437210bb9396471050c",
         tokenId: "2288",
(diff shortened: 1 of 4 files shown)


test/integration/postOrder.spec.ts
```diff
@@ -34,7 +34,7 @@ suite("SDK: order posting", () => {
 
     const sellOrder = {
       accountAddress: walletAddress,
-      startAmount: LISTING_AMOUNT,
+      startAmount: +LISTING_AMOUNT,
```
ryanio, line 37: maybe keep one as a string or it might be smart to test all variations of bignumberish inputs to ensure their resulting values are equal

ryanio (approved):

PR #1032, opened 2023-06-08
send API key header in correct format

cterech:
## Motivation
when we enforced keys on /v1, saw a user note that it broke their code even with a key. This format is what our API expects for the auth header


sambarnes (approved):
👍 

any context around this change? was it recently changed from X-API-KEY?

the old style seems more conventional, havent seen many headers with underscores -- so just curious

cterech:
> 👍
> 
> any context around this change? was it recently changed from X-API-KEY?
> 
> the old style seems more conventional, havent seen many headers with underscores -- so just curious

@sambarnes agreed this is uglier - i haven't actually tested it yet, maybe will just update the API to accept different formats instead

PR #1055, opened 2023-06-29
Add 4 new endpoints and support multichain

thenerdassassin:
## Motivation

The API has added 4 new multichain endpoints which are being implemented here:

- List NFTS by Collection
- List NFTS by Contract
- Get an NFT
- Refresh Metadata for an NFT


src/constants.ts
```diff
@@ -8,18 +6,7 @@ export const API_BASE_MAINNET = "https://api.opensea.io";
 export const API_BASE_TESTNET = "https://testnets-api.opensea.io";
 export const API_PATH = `/api/v${API_VERSION}`;
 
-export const MERKLE_VALIDATOR_MAINNET =
```
thenerdassassin, line 11: I couldn't find this used anywhere

src/constants.ts
```diff
@@ -8,18 +6,7 @@ export const API_BASE_MAINNET = "https://api.opensea.io";
...
-export const WETH_ADDRESS_BY_NETWORK = {
-  [Chain.Mainnet]: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
-  [Chain.Goerli]: "0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6",
-} as const;
-
-export const DEFAULT_ZONE_BY_NETWORK = {
-  [Chain.Mainnet]: "0x0000000000000000000000000000000000000000",
-  [Chain.Goerli]: "0x0000000000000000000000000000000000000000",
-} as const;
+export const DEFAULT_ZONE = "0x0000000000000000000000000000000000000000";
```
thenerdassassin, line 9: Do we want the zero address to be the default? If we do, should it be for all of them?

src/sdk.ts
```diff
@@ -439,6 +432,7 @@ export class OpenSeaSDK {
     paymentTokenAddress?: string;
     buyerAddress?: string;
   }): Promise<OrderV2> {
+    //TODO: Make this function multichain compatible
```
thenerdassassin, line 435: Making changes to the createBuy/createSellOrder to support multichain will have cascading changes so holding off until a separate PR

cterech (approved):
lgtm - should we add methods in top level ```sdk.ts``` for listing / getting nfts?

src/constants.ts
ryanio: you can use `ethers.constants.AddressZero`

src/api/api.ts
```diff
@@ -269,6 +284,97 @@ export class OpenSeaAPI {
...
+        getListNFTsByCollectionPath(slug),
+        {
+          limit,
+          next,
+        }
+      );
+    } catch (error) {
+      _throwOrContinue(error, retries);
+      await delay(1000);
+      return this.getNFTsByCollection(slug, limit, next, retries - 1);
```
ryanio, line 312: instead of putting this retry logic everywhere you can possibly just use poll which has expontential backoff logic https://docs.ethers.org/v5/api/utils/web/#utils-poll

src/sdk.ts
```diff
@@ -339,11 +332,11 @@ export class OpenSeaSDK {
     expirationTime?: BigNumberish;
     paymentTokenAddress?: string;
   }): Promise<OrderV2> {
+    //TODO: Make this function multichain compatible
```
ryanio, line 335: what happens if the fn is passed with a chain like Klaytn? should we throw an error here?

src/api/api.ts
```diff
@@ -75,9 +83,16 @@ export class OpenSeaAPI {
 
     switch (config.chain) {
       case Chain.Goerli:
+      case Chain.Sepolia:
```
ryanio, line 86: possibly makes sense to add a `is_testnet(chain: Chain)` helper and just use that here, then we can use the helper fn elsewhere in the future as well

src/sdk.ts
thenerdassassin: This is calling the V1 API still so you would likely get some sort of not found exception. But we can check for Goeril or Mainnet otherwise throw an error

ryanio (approved):

src/types.ts
```diff
@@ -87,8 +87,27 @@ export interface OpenSeaAPIConfig {
...
+
+  //Testnet Chains
   Goerli = "goerli",
+  Sepolia = "sepolia",
+  Mumbai = "mumbai",
+  Baobab = "baobab",
+  BNBTestnet = "bsctestnet",
+  ArbitrumGoerli = "arbitrum_goerli",
+  Fuji = "avalanche_fuji",
+  OptimisGoerli = "optimism_goerli",
```
ryanio, line 109: typo
```suggestion
  OptimismGoerli = "optimism_goerli",
```

src/types.ts
```diff
@@ -87,8 +87,27 @@ export interface OpenSeaAPIConfig {
...
+  Polygon = "matic",
+  Klaytn = "klaytn",
+  BNB = "bsc",
+  Arbitrum = "arbitrum",
+  ArbitrumNova = "arbitrum_nova",
+  Avalanche = "avalanche",
+  Optimism = "optimism",
+  Solana = "solana",
+
+  //Testnet Chains
```
ryanio, line 101:
```suggestion
  // Testnet Chains
```

PR #1058, opened 2023-06-30
Use zone returned by buildOffer when creating collection offers.

thenerdassassin:
## Motivation

Using the default zone (zero address) was causing the collection offer orders to fail.

## Solution

Instead, use the zone returned by the builder endpoint.


sunwrobert (approved):
lgtm. we actually expose a `offer_protection_enabled` flag that we could add in a future version so people can choose whether or not they want to use the zone. can be done later though

PR #1074, opened 2023-07-08
feat: add lower marketplace fees option.

FiveNine:
## Motivation
I was looking in the SDK and apparently it is hardcoded to use the 2.5% marketplace fees whereas there seems to be an option in the API to utilize lower marketplace fees (0.5% and 0%). 

## Solution
Added an optional boolean to apply lower marketplace fees as needed.


ascende-dev:
Would love to see this added. @thenerdassassin mentioned in Discord that it should be possible to submit listings using the low marketplace fee: 0% if mandatory creator fees, otherwise 0.5%: 
 https://discord.com/channels/397566282729390110/434175585745633280/1121935135193976972

ryanio:
Thanks for the PR!

Nit could you capitalize it Marketplace instead of MarketPlace?

Also could you add some unit tests to ensure the outcomes are proper?

Thank you!

FiveNine:
I added some tests. (first time writing unit tests)
let me know if there's something wrong with the way i've made the tests.

FiveNine:
On another note, I left the `createCollectionOffer` to not utilize the low marketplace fee as I wasn't sure if that's a thing in the API or opensea pro in general.

ryanio:
thanks @FiveNine! looks great. i was discussing with @thenerdassassin and we might not even need a param for this and just automatically set the lower marketplace fee. I'll try to open a PR to merge into yours today and we can get this in, so no need to update your branch further for now. i'd also prefer if these tests would be unit tests rather than integration but I'll see how feasible that is.

FiveNine:
so you're going to permanently set the `opensea_fees` to 0% or 0.5% (depending on creator fees) on the API? Do you guys have a timeline on when i can expect this to work? O.O
TThanks for reviewing :)

ascende-dev:
Being able to use the lower marketplace fees with OpenSea-js will make a big difference for us. We will be listing dozens of NFTs daily when we can use it. We'd like to help build inventory and structure in the market for OpenSea 😀

ascende-dev:
> actually thinking about this further, it feels like we should just return the proper value in the api opensea_fees response, then we wouldn't need any changes in the sdk and wouldn't get out of sync in the future

@ryanio really like this idea about returning the value in the opensea_fees response

We're working on two tools right now:

* One holds marketplace grid positions sorted low price, like: 21st, 24th, 27th. With each update it adjusts to keep the position
* The other will decline the price over time in steps with each update

The low marketplace fee will help us run these tools

ascende-dev:
@ryanio Checking in to see if you have progress update about the opensea_fees. We have our script working very well on Testnet that updates prices to target certain positions in the marketplace grid like 10th, 15th, 20th. We'd would like to move to Mainnet. Having the correct lower fees will help us be able to list with OpenSea-js

jalagar (approved):

Fairy4422 (approved):

ascende-dev:
Checking in on this PR. Looks like we got 1 approval. Is there is anything else needed for this to get merged?

ryanio:
sorry for the delay, yes still working on some internal logic but another week or two and we should be able to get this resolved without needing this PR by returning opensea_fees field properly, will keep this PR open meanwhile

ascende-dev:
hi @ryanio, wanted to check in with you and see if you and the team had accomplished what you wanted in the last comment

ascende-dev:
@ryanio checking back, does the opensea_fees allow for the lower OpenSea Pro pricing when making listings?

ryanio:
included in #1204

ryanio:
will have an update here shortly

ryanio:
@FiveNine @ascende-dev thanks for your patience. the lower marketplace fee was a temporary promotion, the 2.5% fee is required again for all OpenSea orders, so you can use the "fees" response from the api to format the opensea fees and creator fees necessary for an order for a collection. More info here: https://docs.opensea.io/changelog/opensea-fees

Issue #1076, opened 2023-07-09
opensea-js v6.1.1 does not allow me to create orders on MATIC. Any ETA?


munimdev:
Tried the new version to crate an offer, it's throwing this error `Error: Creating orders on matic not yet supported by the SDK.`

My code:
```javascript
const { ethers } = require("ethers");
const { OpenSeaSDK, Chain } = require("opensea-js");

// This example provider won't let you make transactions, only read-only calls:
const provider = new ethers.providers.JsonRpcProvider(
  "https://mainnet.infura.io"
);

const wallet = new ethers.Wallet(
  "<key>",
  provider
);

const openseaSDK = new OpenSeaSDK(provider, {
  chain: Chain.Polygon,
  apiKey: "<API>",
});

// Token ID and smart contract address for a non-fungible token:
const { tokenId, tokenAddress } = {
  tokenId: "<id>",
  tokenAddress: "<address>",
};
// The offerer's wallet address:
const accountAddress = "<account address>";

const main = async () => {
  const offer = await openseaSDK.createBuyOrder({
    asset: {
      tokenId,
      tokenAddress,
    },
    accountAddress,
    // Value of the offer, in units of the payment token (or wrapped ETH if none is specified):
    startAmount: 1.2,
  });

  console.log("Offer created: ", offer);
};

main();

```

Any ETA on when we can expect to create offers? And is there currently a way to do this with the API? A working example would be very much appreciated.

_Originally posted by @munimdev in https://github.com/ProjectOpenSea/opensea-js/issues/1062#issuecomment-1626935382_

ryanio:
Are you sure you're on v6.1.1? I would delete your node_modules and reinstall. What happens if you execute the cmd `cat node_modules/opensea-js/package.json`

ryanio:
You also have to pass the wallet to the 4th constructor arg of the library

munimdev:
> Are you sure you're on v6.1.1? I would delete your node_modules and reinstall. What happens if you execute the cmd `cat node_modules/opensea-js/package.json`

I'm on Windows but here's the contents of that file:

```json
{
  "name": "opensea-js",
  "version": "6.1.1",
  "description": "JavaScript SDK for the OpenSea marketplace helps developers build new experiences using NFTs and our marketplace data!",
  "license": "MIT",
  "author": "OpenSea Developers",
  "homepage": "https://docs.opensea.io/reference/api-overview",
  "repository": {
    "type": "git",
[...]

PR #1077, opened 2023-07-10
Remove multichain block on buyOrder

thenerdassassin:
## Motivation

Fixes issue #1076 

## Solution

Remove clause blocking multichain buy orders.


munimdev:
Hi, could you write some tests as well to make sure this works with the newly added chains such as Polygon etc?

ryanio (approved):

Issue #1064, opened 2023-07-06
Adding quantity to the orders retrieval or another way of getting Unit Price from listings


UTkzhang:
Just checking out the SDK and can't find a way to easily see the current listing unit price (lowest) of a given NFT asset. The quote comes back with currentPrice which is the total price, but no easy way to retrieve quantity. Some guidance here would be appreicated!

Thanks

ryanio:
seems reasonable, would you be interested in working on a PR to add this? we are open to contributions!

UTkzhang:
I will take a look today

UTkzhang:
I will take a look today

UTkzhang:
@ryanio 

Something strange is happening when I try to build and npm link locally. With the latest installed package, everything works fine on my test client. However, when I use link, i get Either a provider or custom signer with provider must be provided. This is thrown by 

 this.seaport_v1_5 = new Seaport(this._signerOrProvider, {
      overrides: { defaultConduitKey: OPENSEA_CONDUIT_KEY },
    });
    
I checked this line in the npm package as well as my local build and cannot find anything different about it, the underlying dep import { Seaport } from "@opensea/seaport-js" is the exact same version. Have you any ideas re. this?

ryanio:
I've experienced that before doing local development it is quite strange. Can you log _providerOrSigner between the two and see how they compare?

UTkzhang:
They are the same

UTkzhang:
logic is very straightforward but i am unable to test locally, please take a look at the linked PR.

PR #1086, opened 2023-07-12
expose new field for remaining quantity

UTkzhang:
## Motivation

We wanted to get unit price of the lowest listing (aka a "quote" of a given NFT asset from Opensea) and needed an easier way to access the quantity field for a given order, as the current price is given as a total rather than a unit price. 

For more details see: https://github.com/ProjectOpenSea/opensea-js/issues/1064

## Solution
expose the "remaining_quantity" field from API.

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -93,7 +93,7 @@ export class OpenSeaSDK {
 
     this.provider = provider;
     this._signerOrProvider = wallet ?? this.provider;
-
+    console.log(this._signerOrProvider);
     this.seaport_v1_5 = new Seaport(this._signerOrProvider, {
       overrides: { defaultConduitKey: OPENSEA_CONDUIT_KEY },
     });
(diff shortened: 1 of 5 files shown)


src/sdk.ts
```diff
@@ -93,7 +93,7 @@ export class OpenSeaSDK {
 
     this.provider = provider;
     this._signerOrProvider = wallet ?? this.provider;
-
+    console.log(this._signerOrProvider);
```
thenerdassassin, line 96: Can you remove this?

UTkzhang: yes

thenerdassassin (commented):
If you would like this to be published now, increment the patch version so it can be published.

UTkzhang:
ok sure

UTkzhang:
would it be 6.1.3?

UTkzhang:
are these tests a problem?

thenerdassassin (approved):

Issue #1092, opened 2023-07-14
Change chain in testnet


tienvv1234:
Can I change the chain in the testnet environment? The document only talks about the Goerli chain. I tried to modify the code, but it doesn't seem to be working, It's only work with Goerli

```
const ethers = require('ethers');
const { OpenSeaSDK, Chain } = require('opensea-js');
const provider = new ethers.providers.JsonRpcProvider(
    'https://api-testnet.polygonscan.com/'
);
const openSea = new OpenSeaSDK(
    provider,
    {
        chain: Chain.Mumbai,
    },
);

let a = await openSea.api.getAssets({
            owner: '0x6E82A097E79251468E54dbE7FE6c943367d96EF0',
        });
```
it return not thing to me
```
result { assets: [], next: null, previous: null, estimatedCount: undefined }
```
I've already created a assets in testnet (https://testnets.opensea.io/assets/mumbai/0x2953399124f0cbb46d2cbacd8a89cf0599974963/41523119144019418335847972109562421607287855164231022950688170492050305712129)
![image](https://github.com/ProjectOpenSea/opensea-js/assets/34591094/94968870-12fa-4b6c-af6e-9ea684cb678e)
Can you guys help me understand how to change the chain, or what did I do wrong?

ryanio:
getAssets is a v1 api that is not multichain, please use getNFTs

tienvv1234:
Thank you it's working

ryanio:
Great thanks for letting me know. I'll open a PR tomorrow to throw an error when using the v1 methods multichain and suggest to use getNFTs to help others 👍

tienvv1234:
Thank you it will be useful

tienvv1234:
@ryanio how can I check the Balance
here is the code I've written following the SDK document 
```
 a = (await openSea.api.getNFTsByCollection('mumbai-collection-10')).nfts;
        console.log('a', a);

        const asset = {
            tokenAddress: a[0].contract, // CryptoKitties
            tokenId: a[0].identifier,
            tokenStandard: 'ERC1155',
        };
        console.log('asset', asset);

        const balance = await openSea.getBalance({
            accountAddress: '0x6E82A097E79251468E54dbE7FE6c943367d96EF0', // string
            asset, // Asset
        });
 ```
 but I it will return to me a error 
 ```
 TypeError: Cannot read properties of undefined (reading 'balanceOf')
[...]

PR #1095, opened 2023-07-14
throw error when using v1 apis for multichain

ryanio:
Fixes user encountered issues, https://github.com/ProjectOpenSea/opensea-js/issues/1092


thenerdassassin (commented):
I think we are missing getBundle and getBundles

thenerdassassin (approved):

ryanio:
> I think we are missing getBundle and getBundles

i tested and these seem to be working multichain

PR #1110, opened 2023-07-19
add collection offer reads and token id decoding

cterech:
## Motivation

We added a new field to our offers APIs which return an encoded list of token ids applicable for a given criteria so users can decipher what tokens are in an offer

Testing:
```
  SDK: getCollectionOffers
    ✔ Get Collection Offers (337ms)
```


src/utils/utils.ts
```diff
@@ -426,3 +426,36 @@ export const isValidProtocol = (protocolAddress: string): boolean => {
   );
   return validProtocolAddresses.includes(checkSumAddress);
 };
+
+/**
+ * Decode an encoded list of token ids
+ * @param encodedTokenIds encoded token ids from the criteria of an order
+ * @returns a list of token ids
+ */
+export const decodeTokenIds = (encodedTokenIds: string): string[] => {
```
cterech, line 435: chatgpt generated

ryanio: nice. in the typedoc could you give a little explanation of how it is encoded to be decoded, so a succinct summary is understandable without reading into the code

src/utils/utils.ts
```diff
@@ -426,3 +426,36 @@ export const isValidProtocol = (protocolAddress: string): boolean => {
...
+  for (const r of ranges) {
+    if (r.includes(":")) {
+      const [start, end] = r.split(":").map(Number);
+      numbers.push(
+        ...Array.from({ length: end - start + 1 }, (_, i) =>
+          (start + i).toString(),
+        ),
+      );
+    } else {
+      numbers.push(Number(r).toString());
```
ryanio, line 456: why convert to number then back to string?

src/utils/utils.ts
cterech: will do! planning on adding it in docs when we release

src/utils/utils.ts
```diff
@@ -426,3 +426,36 @@ export const isValidProtocol = (protocolAddress: string): boolean => {
...
+  }
+
+  const ranges = encodedTokenIds.split(",");
+  const numbers: string[] = [];
+
+  for (const r of ranges) {
+    if (r.includes(":")) {
+      const [start, end] = r.split(":").map(Number);
+      numbers.push(
+        ...Array.from({ length: end - start + 1 }, (_, i) =>
```
ryanio, line 451: nice es6 solution, although would recommend we make a little util fn and import, then it should be really nice and concise to read here

```typescript
const range = (start: number, end: number) => Array.from(
  Array(Math.abs(end - start) + 1), 
  (_, i) => start + i
);
```

test/integration/getCollectionOffers.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    assert(response, "Response should not be null");
+    assert(response.offers, "Collection offers should not be null");
+    assert(response.offers.length > 0, "Collection offers should not be empty");
+    const offer = response.offers[0];
+    assert(offer.order_hash, "Order hash should not be null");
+    const tokens = offer.criteria.encoded_token_ids;
+    assert(tokens, "Criteria should not be null");
+
+    const encodedTokends = offer.criteria.encoded_token_ids;
```
ryanio, line 19: typo `encodedTokends`

ryanio:
could you add a few unit tests for `decodeTokenIds` to test that every scenario *, range, etc. returns the right output? or just a few complicated examples that output to what we expect

cterech:
> could you add a few unit tests for `decodeTokenIds` to test that every scenario *, range, etc. returns the right output? or just a few complicated examples that output to what we expect

yup good call!

src/utils/utils.ts
ryanio: actually this is not going to work for big numbers, you'll have to use BigNumber in case the token ids are past the max safe js number

ryanio: you'll want something like this

```typescript
const [start, end] = r.split(":").map(BigNumber.from);
```

```typescript
const range = (start: BigNumber, end: BigNumber) => Array.from(
  Array(end.sub(start).add(1).toNumber()), 
  (_, i) => start.add(i).toNumber()
);
```

ryanio:
also would be nice to bump the version to 6.1.5 then we can release after we merge this in

.nvmrc
```diff
@@ -1 +1 @@
-18.17.0
+18.17.0
```
thenerdassassin, line 1: ?

cterech: not sure why this is in here - can't seem to get it out of the diff

src/api/api.ts
```diff
@@ -221,20 +223,35 @@ export class OpenSeaAPI {
     return response;
   }
 
+  /**
+   * Post collection offer
```
thenerdassassin, line 227: This is not the right documentation....

Also can you make this follow the typedoc standard with @param @return, etc

src/utils/utils.ts
```diff
@@ -420,3 +420,71 @@ export const isValidProtocol = (protocolAddress: string): boolean => {
...
+  for (const range of ranges) {
+    if (range.includes(":")) {
+      const [startStr, endStr] = range.split(":");
+      const start = BigNumber.from(startStr);
+      const end = BigNumber.from(endStr);
+      const diff = end.sub(start).add(1);
+
+      if (diff.isNegative() || diff.isZero()) {
+        throw new Error(
+          "Invalid range. End value must be greater than or equal to the start value.",
```
thenerdassassin, line 476: Might be helpful to include the start, end values here for debugging

src/api/api.ts
```diff
@@ -221,20 +223,35 @@ export class OpenSeaAPI {
...
+
   /**
    * Post collection offer
    */
   public async postCollectionOffer(
     order: ProtocolData,
     slug: string,
     retries = 0,
-  ): Promise<PostOfferResponse | null> {
+  ): Promise<Offer | null> {
```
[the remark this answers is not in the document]
cterech: fixed

src/utils/utils.ts
ryanio: agreed, you can also just do `if (diff.lte(0))`

cterech: good idea, fixed

src/utils/utils.ts
```diff
@@ -420,3 +420,71 @@ export const isValidProtocol = (protocolAddress: string): boolean => {
...
+        );
+      }
+
+      for (let i = 0; i < diff.toNumber(); i++) {
```
ryanio, line 480: we still have to use BigNumber here, make i a bignumber and do i.lt(diff), i.add(1)

src/api/api.ts
```diff
@@ -222,19 +224,45 @@ export class OpenSeaAPI {
   }
 
   /**
-   * Post collection offer
+   * Get collection offers for a given slug in the API
+   *
+   * @param slug The collection you would like to list offers for
+   * @param retries Number of times to retry if the service is unavailable for any reason
+   *
+   * @returns ListCollectionOffersResponse
```
thenerdassassin, line 232: Can you do a {@link ...} to this object?

src/utils/utils.ts
cterech: fixed

src/api/api.ts
cterech: done

src/api/api.ts
```diff
@@ -222,19 +224,45 @@ export class OpenSeaAPI {
...
+  }
+
+  /**
+   * Post a collection offer to the API
+   *
+   * @param order The order to post
+   * @param slug The collection you would like to post an offer for
+   * @param retries Number of times to retry if the service is unavailable for any reason
+   *
+   * @returns Offer
```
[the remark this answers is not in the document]
cterech: done

src/api/api.ts
```diff
@@ -222,19 +224,45 @@ export class OpenSeaAPI {
...
+  }
+
+  /**
+   * Post a collection offer to the API
+   *
+   * @param order The order to post
+   * @param slug The collection you would like to post an offer for
+   * @param retries Number of times to retry if the service is unavailable for any reason
+   *
+   * @returns (@link Offer}
```
thenerdassassin, line 256: mistype `(` > `{`

ryanio (approved):

Issue #1109, opened 2023-07-19
how to listing assets to mumbai chain


tienvv1234:
When listing assets on the website https://testnets.opensea.io/, I have the option to choose the currency for the price, and I want to select the Matic. However, when I try calling the createSellOrder function with the SDK, it always lists the assets on the Goerli chain.
![image](https://github.com/ProjectOpenSea/opensea-js/assets/34591094/1662d278-87e0-4700-a9c8-8714cec7d5f1)
How can I listing on Mumbai chain

thenerdassassin:
Can you send over what your sdk initialization looks like?

tienvv1234:
Hi @thenerdassassin sorry I was wrong, it can create sell in Mumbai chain, the problem is I can not fulfillOrder with my asset
here is my code
```
const ethers = require('ethers');
const { OpenSeaSDK, Chain, EventType } = require('opensea-js');
const { handleSDKEvents } = require('./event-listenter');
const provider = new ethers.providers.JsonRpcProvider(
    'https://polygon-mumbai.infura.io/v3/4458cf4d1689497b9a38b1d6bbf05e78'
);
const openSea = new OpenSeaSDK(
    provider,
    {
        chain: Chain.Mumbai,
        // apiKey: '2fce588f0e1c48e2825cc68c25058665',
    },
    console.log(OpenSeaSDK._signerOrProvider),
    new ethers.Wallet(
        privateKey,
        provider
    )
);

const asset = {
            tokenAddress: '0x2953399124F0cBB46d2CbACD8A89cF0599974963', // CryptoKitties
            tokenId: '41523119144019418335847972109562421607287855164231022950688170494249328967681',
            tokenStandard: 'ERC1155',
        };
        console.log('asset', asset);
        // const expirationTime = Math.round(Date.now() / 1000 + 60 * 60 * 24);

        
        const balance = await openSea.getBalance({
            accountAddress: acc1, // string
            asset, // Asset
        });

        console.log('balance', balance);
        
        const ownsKitty = balance.gt(0);
        console.log('ownsKitty', ownsKitty);

        if (!ownsKitty) {
            throw new Error('You do not own this asset!');
        }
        
        const listings = await openSea.api.getOrder({
            side: 'ask',
            assetContractAddress: asset.tokenAddress,
            tokenId: asset.tokenId,
        });
        // console.log('listings', JSON.stringify(listings, null, 2));

        // Transaction created buying items
[...]

PR #1111, opened 2023-07-19
add addresses for polygon (matic/mumbai) and klaytn shared storefront

ryanio:
fixes https://github.com/ProjectOpenSea/opensea-js/issues/1109#issuecomment-1642311112

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -10,9 +10,24 @@ export const API_PATH = `/api/v${API_VERSION}`;
 
 export const DEFAULT_ZONE = ethers.constants.AddressZero;
 
+// Ignore eslint no-unused-modules for below to keep backward compatibility
+// in case a downstream user was already using these imports directly.
+// These can be made non-exported in next major-versioned release.
+// eslint-disable-next-line import/no-unused-modules
 export const SHARED_STOREFRONT_ADDRESS_MAINNET =
   "0x495f947276749ce646f68ac8c248420045cb7b5e";
+// eslint-disable-next-line import/no-unused-modules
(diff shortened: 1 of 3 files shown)


src/constants.ts
```diff
@@ -10,9 +10,24 @@ export const API_PATH = `/api/v${API_VERSION}`;
...
+const SHARED_STOREFRONT_ADDRESS_POLYGON =
+  "0x2953399124f0cbb46d2cbacd8a89cf0599974963";
+const SHARED_STOREFRONT_ADDRESS_KLAYTN =
+  "0x5bc519d852f7ca2c8cf2d095299d5bb2d13f02c9";
+export const SHARED_STOREFRONT_ADDRESSES = [
+  SHARED_STOREFRONT_ADDRESS_MAINNET,
+  SHARED_STOREFRONT_ADDRESS_GOERLI,
+  SHARED_STOREFRONT_ADDRESS_POLYGON,
+  SHARED_STOREFRONT_ADDRESS_KLAYTN,
+].map((address) => address.toLowerCase());
```
sambarnes, line 31: any particular reason for moving from set to array here? would think the uniqueness properties of a set would be nice -- but maybe im just thinking in python

sambarnes (approved):

src/constants.ts
ryanio: yeah, just felt unneeded really, you are technically correct but since we are defining it ourselves here with no user input i feel like it's no biggie

sambarnes: sounds good!

PR #1133, opened 2023-08-04
feat: add base l2 chain identifiers

sambarnes:
## Motivation

base went live yesterday, adding here & will update the docs


BCLeFevre (approved):

ryanio:
can you also add to `isTestChain` please

sambarnes:
updated the supported chains docs too now: https://docs.opensea.io/reference/supported-chains

PR #1140, opened 2023-08-07
Simplify isValidProtocol check

ryanio:
- DRY `isValidProtocol` by introducing `requireValidProtocol` that throws the consistent error
- Rename `_checkAccountIsAvailable` to `_requireAccountIsAvailable` to hint that an error will be thrown


sambarnes (approved):
🔥

PR #1148, opened 2023-08-09
add getNFTsByAccount

ryanio:
Adds new v2 API endpoint: https://docs.opensea.io/reference/retrieve-nfts-by-account

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -430,7 +430,7 @@ export class OpenSeaAPI {
     limit: number | undefined = undefined,
     next: string | undefined = undefined,
     retries = 1,
-    { chain = this.chain }: { chain?: Chain } = { chain: this.chain },
+    chain = this.chain,
   ): Promise<ListNFTsResponse> {
     let response;
     try {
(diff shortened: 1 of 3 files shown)


src/api/api.ts
```diff
@@ -416,6 +417,41 @@ export class OpenSeaAPI {
...
+   * @param next Cursor to retrieve the next page of NFTs
+   * @param retries Number of times to retry if the service is unavailable for any reason.
+   * @returns The {@link ListNFTsResponse} returned by the API.
+   */
+  public async getNFTsByAccount(
+    address: string,
+    limit: number | undefined = undefined,
+    next: string | undefined = undefined,
+    retries = 1,
+    { chain = this.chain }: { chain?: Chain } = { chain: this.chain },
```
thenerdassassin, line 433: Missing this param in the documentation.
Does this need to be a type? why not `chain: Chain | undefined`?

ryanio: good call, fixed (doing `chain = this.chain` makes the type `chain?: Chain` which accepts undefined btw)

thenerdassassin (approved):

PR #1154, opened 2023-08-13
Fix getting extraData in order fulfillment

### LINKED ISSUE #1153: Erro when uising fulfillOrder()
### Component

API

### Have you ensured that all of these are up to date?

- [X] opensea-js
- [X] Node (minimum v16)

### What version of opensea-js are you on?

6.1.7

### What function is the bug in?

fulfillOrder

### Operating System

macOS (Apple Silicon)

### Describe the bug

I get a order data from the code below
`return await seaport.api.getOrder({
        side: side,
        orderDirection: orderDirection,
        orderBy: orderBy,
        // limit: limit,
        assetContractAddress: assetContractAddress,
        tokenId: tokenId,
        // maker: maker
    })`

Then, I pass the order data to fulfillOrder method.
`    return await seaport.fulfillOrder({
        order: order,
        accountAddress: publicKey,
    });`

But, the error shows: 
` const fulfillmentOrder = result.fulfillment_data.transaction.input_data.orders[0]; TypeError: Cannot read properties of undefined (reading '0')`

Then, I thought if 'order' should be an array? so, I modified the code below:
`    return await seaport.fulfillOrder({
        order: [].push(order),
        accountAddress: publicKey,
    });`

Finally, the error shows here:
`Error: invalid address (argument="address", value=undefined, code=INVALID_ARGUMENT, version=address/5.7.0)`

Any suggestions for that? Thanks in advance

ryanio:
extraData is only provided in an array of Advanced Orders, improves the guard

Fixes #1153

TODO need to add a local hardhat forked setup for an integration test to get varying valid orders (basic, advanced with protection, etc.) and fulfill them on the fork and ensure the tx is successful. might try to add this today.

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -705,10 +705,9 @@ export class OpenSeaSDK {
 
       // If the order is using offer protection, the extraData
       // must be included with the order to successfully fulfill.
-      const fulfillmentOrder =
-        result.fulfillment_data.transaction.input_data.orders[0];
-      if ("extraData" in fulfillmentOrder) {
-        extraData = (fulfillmentOrder as AdvancedOrder).extraData;
+      const inputData = result.fulfillment_data.transaction.input_data;
+      if ("orders" in inputData && "extraData" in inputData.orders[0]) {
+        extraData = (inputData.orders[0] as AdvancedOrder).extraData;
       }
 
       const signature = result.fulfillment_data.orders[0].signature;


src/sdk.ts
```diff
@@ -705,10 +705,9 @@ export class OpenSeaSDK {
 
       // If the order is using offer protection, the extraData
       // must be included with the order to successfully fulfill.
-      const fulfillmentOrder =
-        result.fulfillment_data.transaction.input_data.orders[0];
-      if ("extraData" in fulfillmentOrder) {
-        extraData = (fulfillmentOrder as AdvancedOrder).extraData;
+      const inputData = result.fulfillment_data.transaction.input_data;
```
kaiqiangh, line 708: for my understanding, line 708 should be: 

`const fulfillmentOrder = result.fulfillment_data.orders[0];`

Can you recheck it? btw, I have no idea how to test the new fix on my side. It looks like you modified ts file and it should be built it locally. I do not know how to using the modified version on my side. Thanks

ryanio: i believe the orders object doesn't contain the AdvancedOrder extraData, so we have to look at input_data. you can try modifying the changed lines in your local built js files, there shouldn't be any ts issues with the changed lines.

thenerdassassin (approved):

PR #1162, opened 2023-08-16
Add mumbai WETH address

mohanv-inmar:
## Motivation
Added Mumbai WETH address, So that English auctions can be created over `mumbai` chain.

## Solution
Adding Mumbai WETH token address to `getWETHAddress` function


ryanio:
LGTM, thanks for opening! Can you move Mumbai to be under Polygon? Thank you

mohanv-inmar:
> LGTM, thanks for opening! Can you move Mumbai to be under Polygon? Thank you

Yeah, I've moved it.

PR #1198, opened 2023-09-19
Fix _requireAccountIsAvailable, replace instanceof

nickfujita:
## Motivation

In projects that are using esm modules instead of commonjs, `instanceof` does not return true when one instance of `Wallet` in client project is using `webpack://<project_name>/node_modules/@ethersproject/wallet/lib.esm/index.js` instead of `/node_modules/@ethersproject/wallet/lib/index.js`. While they are the same class from the same package, the js `instanceof` operation does not recognize them as matching.

## Solution

- Matches the name of the instantiation class of the `Wallet` instance passed into `OpenSeaSDK` constructor, and the `Wallet` class required inside opensea-js
- Casts instance to proper type before comparing address

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -1084,12 +1084,16 @@ export class OpenSeaSDK {
   private async _requireAccountIsAvailable(accountAddress: string) {
     const accountAddressChecksummed = ethers.utils.getAddress(accountAddress);
     if (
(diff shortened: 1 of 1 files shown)


ryanio:
Thank you for opening a fix! We will review and get this in.

ryanio:
included in #1204

PR #1205, opened 2023-09-26
Remove schedule from renovate.json

ryanio:
Trying to automate automerge workflow, so we don't need to be on any specific schedule since passing PRs should be automerged


ryanio:
i was wrong, going to add this schedule to seaport-js instead lol

PR #1239, opened 2023-10-17
Use GitHub app for token generation

thenerdassassin:
Use GH Application as an installable app to gain write access and push to the developer-docs repo

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: npm
-      - uses: opensea-public-docs-publisher
+      - uses: tibdex/github-app-token@v1
         id: get_token
         with:
           app_id: ${{ secrets.PUBLIC_DOC_PUBLISHER_APP_ID }}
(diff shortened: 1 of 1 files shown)


nsheaps (approved):

ryanio (approved):

zoe-y-zuo (approved):

.github/workflows/docs.yml
```diff
@@ -15,6 +15,11 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: npm
+      - uses: opensea-public-docs-publisher
```
nsheaps, line 18: I don't think this will work, you need to use the action that's specified in the notion doc, sorry I didn't catch this sooner

nsheaps, line 18:
```suggestion
      - uses: tibdex/github-app-token@v1
```

jgriffos2 (approved):

.github/workflows/docs.yml
```diff
@@ -15,6 +15,11 @@ jobs:
...
+        id: get_token
+        with:
+          app_id: 409633
```
nsheaps, line 21: please use the secret here
```suggestion
        app_id: ${{ secrets.PUBLIC_DOC_PUBLISHER_APP_ID }}
```

It's not really a secret, but makes it so if we change to a new app it doesn't need workflow changes

nsheaps (approved):

PR #1332, opened 2023-12-22
Add asset transfer function

ryanio:
Adds transfer() helper for assets

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -190,7 +190,6 @@ export class OpenSeaSDK {
       this._signerOrProvider,
     );
 
-    wethContract.connect(this.provider);
     try {
       const transaction = await wethContract.deposit({ value });
       await this._confirmTransaction(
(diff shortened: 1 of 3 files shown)


src/sdk.ts
```diff
@@ -937,6 +939,101 @@ export class OpenSeaSDK {
...
+          this._signerOrProvider,
+        );
+        transaction = contract.transfer(toAddress, amount, overrides);
+        break;
+      }
+      case TokenStandard.ERC1155: {
+        if (asset.tokenId === undefined || asset.tokenId === null) {
+          throw new Error("Missing ERC1155 tokenId for transfer");
+        }
+        if (!amount) {
```
stuckinaboot, line 984: This will throw on 0 amount. That's correct right?

stuckinaboot (approved):
lgtm, see 1 comment

src/sdk.ts
ryanio: yes, should be good to throw early on this as the contract should revert with 0 amount

PR #1343, opened 2024-01-02
add new api fields

ryanio:
adds new api fields:
- collection.totalSupply
- collection.createdDate
- nft.traits

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -230,7 +230,7 @@ export type Trait = {
   trait_type: string;
   /** A field indicating how to display. None is used for string traits. */
   display_type: TraitDisplayType;
-  /** Display type of trait */
+  /** Ceiling for possible numeric trait values */
   max_value: string;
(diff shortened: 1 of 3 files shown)


src/api/types.ts
```diff
@@ -211,12 +211,41 @@ export type NFT = {
...
+ * Trait type returned by OpenSea API.
+ * @category API Models
+ */
+export type Trait = {
+  /** The name of the trait category (e.g. 'Background') */
+  trait_type: string;
+  /** A field indicating how to display. None is used for string traits. */
+  display_type: TraitDisplayType;
+  /** Display type of trait */
+  max_value: string;
```
stuckinaboot, line 234: just to confirm, is string correct here?

stuckinaboot (approved):

src/api/types.ts
ryanio: yes but the comment is wrong, good catch. has to be string even if it's a number for big numbers or decimals

PR #1432, opened 2024-04-03
use required zone when provided

ryanio:


src/sdk.ts
```diff
@@ -406,10 +411,10 @@ export class OpenSeaSDK {
           expirationTime !== undefined
             ? BigInt(expirationTime).toString()
             : getMaxOrderExpirationTimestamp().toString(),
-        zone: DEFAULT_ZONE,
+        zone,
         domain,
         salt: BigInt(salt ?? 0).toString(),
-        restrictedByZone: false,
+        restrictedByZone: zone !== DEFAULT_ZONE,
```
BCLeFevre, line 417: Will this run into issues with 0x000...000. and our pausable zone?

ryanio: no, default zone is null address. pausable zone is only set internally as an analytics identifier that the order was created by us internally, sdk does not use it

PR #1441, opened 2024-04-14
fix(sdk): allow apiBaseUrl overwrite

Dimfred:
# Bug

OpenSeaSDK does not use the `apiBaseUrl` parameter in the `apiConfig`, instead overwrites it with the default parameters.

The PR allows to again overwrite the base url as the user pleases.


ryanio (approved):
Thanks!

PR #1491, opened 2024-06-14
Use ethers.Signer.getAddress() in OpenSeaSDK._getAvailableAccounts()

RLaursen:
## Motivation

The SDK does not currently work with `ethers.Signer` objects without an `address` property. `ethers.Signer` objects may asynchronously retrieve their own address via the `.getAddress()` method, they do not always have an `address` property.

## Solution

Fallback to calling `.getAddress()` in `OpenSeaSDK._getAvailableAccounts`

## Open Questions

Are the added tests sufficient/desired?

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -1292,6 +1292,8 @@ export class OpenSeaSDK {
         (acct) => acct.address,
       );
       availableAccounts.push(...addresses);
+    } else if ("getAddress" in this._signerOrProvider) {
+      availableAccounts.push(await this._signerOrProvider.getAddress());
     }
 
     return availableAccounts;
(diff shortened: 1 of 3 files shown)


ryanio:
thanks for the PR! do you have an example of instantiating a signer that has getAddress()?
also i don't think we need to modify the tests, just the src you included seems fine

RLaursen:
> thanks for the PR! do you have an example of instantiating a signer that has getAddress()? also i don't think we need to modify the tests, just the src you included seems fine

Here's an example from our SDK :)

https://docs.stardust.gg/typescript-sdk/signers/ethers-v6/sign-transaction

RLaursen:
I added some secrets I used for local testing to my fork, required gha `code-quality / test (pull_request)` should pass now I think

RLaursen:
Is anything else required for this? @ryanio

ryanio:
looks good to me! sorry for the delay, will prepare a release now

PR #1513, opened 2024-07-08
Add sei mainnet and testnet chain

sraturi:

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -123,7 +123,8 @@ export enum Chain {
   Solana = "solana",
   /** Zora */
   Zora = "zora",
-
+  /** Sie */
+  Sei = "Sei",
   // Testnet Chains
   // ⚠️NOTE: When adding to this list, also add to the util function `isTestChain`
   /** Sepolia */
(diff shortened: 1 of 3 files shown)


ryanio:
can you add sei mainnet too? seems like we're missing that

src/types.ts
```diff
@@ -148,6 +148,8 @@ export enum Chain {
   SolanaDevnet = "soldev",
   /** Zora Sepolia */
   ZoraSepolia = "zora_sepolia",
+  /** Sei Testnet */
```
stuckinaboot, line 151: just add package.json version bump as well

stuckinaboot (approved):

src/types.ts
```diff
@@ -123,7 +123,8 @@ export enum Chain {
   Solana = "solana",
   /** Zora */
   Zora = "zora",
-
+  /** Sie */
+  Sei = "Sei",
```
ryanio, line 127:
```suggestion
  /** Sei */
  Sei = "sei",
  
```

Issue #1531, opened 2024-08-01
0% Opensea fee Collection will cause "consideration item with 0 start amount" errors


1220moritz:
### Component

API

### Have you ensured that all of these are up to date?

- [X] opensea-js
- [X] Node (minimum v16)

### What version of opensea-js are you on?

7.1.12

### What function is the bug in?

createListing

### Operating System

Linux

### Describe the bug

Tried to call createListing
```
await this.openseaSDK.createListing({
                asset: asset,
                accountAddress: this.myPublicAddress,
                startAmount: priceInEth.toFixed(6),
                expirationTime: expirationTime,
                excludeOptionalCreatorFees: true
            })
```
with these params:
```
{
  tokenAddress: '0x17798ba794ad94dad0b7d3f4b9bc2e9f6486d4b9',
  tokenId: '6689',
  tokenStandard: 'ERC721'
}  //Asset
0xaaaaa9E861aaaa12BaaB313aaaaa6D8aaD20aaaa //My public address (removed some chars)
0.088800 //price
1722548626 //unix
```

and received the error:
`Server Error: You have attempted to create a order, but you provided an offer item or consideration item with a 0 start amount. Remove this 0-amount item if this was intended`

The error is caused, because the collection (tour-de-berace) has 0% Opensea fees but the fee is still included in consideration.
It works after removing the Opensea fees consideration.

PR #1532, opened 2024-08-02
support for collection with 0% Opensea fee

1220moritz:
## Motivation

Fix the issue where collections with zero fees are not handled correctly, causing unnecessary errors.

## Solution

Filter out zero-value fees and add a conditional check to process fees only when they are greater than zero.

Related issue: [#1531](https://github.com/ProjectOpenSea/opensea-js/issues/1531)


ryanio (approved):
thanks! lgtm

PR #1574, opened 2024-10-24
Release B3

sraturi:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "opensea-js",
-  "version": "7.1.13",
+  "version": "7.1.14",
   "description": "TypeScript SDK for the OpenSea marketplace helps developers build new experiences using NFTs and our marketplace data",
   "license": "MIT",
   "author": "OpenSea Developers",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -39,7 +39,8 @@
   "types": "lib/index.d.ts",
   "dependencies": {
     "@opensea/seaport-js": "^4.0.0",
-    "ethers": "^6.9.0"
+    "ethers": "^6.9.0",
+    "opensea-js": "^7.1.13"
```
jalagar, line 43: why is this added?

jalagar (approved):

package.json
sraturi: based on https://www.notion.so/opensea/EPD-Supporting-a-new-chain-7b4f81c6dd26413eb784db78f181afd0?pvs=4#7a76b961bef74121bf6bfe2b9c086f76

jtaaa (changes requested):
Should not add opensea-js to deps

package.json
jtaaa: need to bump the version above. add to deps here

jtaaa (approved):

its2badwespend:
> 
> 
> ## Motivation
> 
> 
> 
> ## Solution
> 
> 
>

PR #1614, opened 2025-01-28
Updated Price type

crypt0biwan:
Updated Price type (mismatch with API response) and added a few asserts examples for price in "BestOffer"

## Motivation

When implementing "opensea-js" package, I can't use the Price type. The API response doesn't contain a "current" field.

## Solution

Updated the type, removed the "current" level. Also added a few asserts in the unit test to validate the format of the API call.
Also, maybe it's a nice recommendation to include "zod" so you can validate API responses against a schema during tests


Boomtokn (approved):

ryanio (approved):
thanks

PR #1615, opened 2025-02-06
Release Bera chain

sraturi:

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -216,6 +216,8 @@ export const getChainId = (chain: Chain) => {
       return "8333";
     case Chain.B3Sepolia:
       return "1993";
+    case Chain.BeraChain:
+      return "80094";
     default:
       throw new Error(`Unknown chainId for ${chain}`);
   }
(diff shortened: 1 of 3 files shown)


src/utils/utils.ts
```diff
@@ -259,6 +261,9 @@ export const getWETHAddress = (chain: Chain) => {
     case Chain.B3:
     case Chain.B3Sepolia:
       return "0x4200000000000000000000000000000000000006";
+    case Chain.BeraChain:
+      // https://beratrail.io/token/0x2F6F07CDcf3588944Bf4C42aC74ff24bF56e7590
+      return "0x2f6f07cdcf3588944bf4c42ac74ff24bf56e7590";
```
sraturi, line 266: Bera also has WBERA address, should we use wbera instead?

jalagar (approved):

src/utils/utils.ts
ryanio: should be WETH

ryanio: actually my bad it should be WBERA since that is what we will accept for offers on OS

sraturi: yup this is weth

Issue #1668, opened 2025-05-04
Include retry-after info on 429 errors


0xmichalis:
### Component

SDK

### Describe the feature you would like

It does not look like the SDK returns the retry-after header that is returned by the API on 429 errors so SDK users have no way to use a backoff mechanism and implement retries when ratelimited. Ideally the value of the retry-after header is included in the response body when the SDK throws the 429 error to enable this sort of capability for SDK users.

ryanio:
this is a great idea

PR #1671, opened 2025-05-06
Return typed ratelimit error including retry-after header

0xmichalis:
Fixes https://github.com/ProjectOpenSea/opensea-js/issues/1668


src/api/api.ts
```diff
@@ -755,6 +756,19 @@ export class OpenSeaAPI {
 
     const response = await req.send();
     if (!response.ok()) {
+      if (response.statusCode === 599) {
```
0xmichalis, line 759: I could also use the standard 429 status code here to make the clause future-proof.

ryanio:
thanks for this PR! it's much appreciated, and sorry it took me so long to review. i'm concerned it might break people's implementations though by throwing a different error if we release it in a patch version. what is the current error response when you get rate limited? i do think it's useful to return the retryAfter data, but maybe we can just add it to the end of the existing error message and it can be grepped? not ideal, but just one thought.

src/api/api.ts
ryanio: 429 makes sense, not sure if it should be catching 599 for rate limit?

0xmichalis: This is the status code currently returned from the API (likely a proxy under your control sitting in front of the actual API).

ryanio:
sorry for the delay here, finally got some time to get back to this. i asked cursor to come up with a way to do this in a non-breaking way, and it came up with this #1718 - i'll release it shortly let me know how it works for you! thanks

PR #1695, opened 2025-06-17
Add Shape to Chain enum

jeffdoubleyou:
## Motivation

Add support for Shape Network

Official Website: https://shape.network/ 

Token: ETH
 

Shape is a creator-focused EVM-equivalent blockchain. Shape creators receive back a portion of the gas spent by users using a concept called “[gasback](https://docs.shape.network/documentation/building-on-shape/gasback).”

 

The Shape network is compatible with EVM wallets, and uses ETH bridged to Shape.

## Solution

Add Shape and testnet to Chain as well as util functions getChainId and getWETHAddress

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -169,8 +169,6 @@ export enum Chain {
   SaigonTestnet = "saigon_testnet",
   /** Abstract Testnet */
   AbstractTestnet = "abstract_testnet",
-  /** Shape Testnet */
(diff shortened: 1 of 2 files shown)


ryanio:
thank you for the PR! didnt realize we missed adding shape

ryanio:
it actually doesnt look like we have shape testnet on https://docs.opensea.io/reference/get_nft, removing

PR #1752, opened 2025-09-25
Minor internal api cleanups to improve comments and avoid errors

0xtygra:
## Motivation

For a personal project I needed to generate the calldata for (but not actually fulfill) a listed NFT. The SDK has an internal api.ts that supports this but from my brief searching it looks like it isn't exposed.

## Solution

image

I ended up forking the repo and using it locally. In doing so, i found that the JSDoc for `api.getOrder` and `api.getOrders` had an `options.limit` that was in fact omitted in the typescript. Additionally, passing an `orderBy: "eth_price"` to `api.getOrder` throws an error, so I also removed that as a configurable option.

A couple of other things i noticed:
- a .env.example would be great, as well as updating the contributing.md to call out the need for a .env to run the tests. In addition, unless it's already done so, exposing the calldata generation functions from api.ts on the exported OpenSeaSDK object would be great!

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -120,8 +120,6 @@ export class OpenSeaAPI {
    * @param options.side The side of the order (listing or offer)
    * @param options.protocol The protocol, typically seaport, to query orders for
    * @param options.orderDirection The direction to sort the orders
-   * @param options.orderBy The field to sort the orders by
-   * @param options.limit The number of orders to retrieve
    * @param options.maker Filter by the wallet address of the order maker
    * @param options.taker Filter by  wallet address of the order taker
    * @param options.asset_contract_address Address of the NFT's contract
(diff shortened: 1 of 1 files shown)


ryanio:
here is what the docs say, maybe we should just update the typedoc to include this?
> The order in which to sort the results. Default: created_date NOTE: If eth_price is used, asset_contract_address and token_id are required.
https://docs.opensea.io/reference/get_listings

0xtygra:
@ryanio i can update the PR to do that! i think i understand better now what the endpoint is for, i realise that each token might have multiple listings (in my mind i was only thinking of the "active" ie cheapest listing and wanted an endpoint to get me the listings of a collection, sorted by eth price such that it gives the "sort by price" view on the collection page)

ryanio:
yeah it's a bit confusing. we used to have multiple listing currencies in OS1 but now in OS2 every chain has one listing currency (usually native, except for polygon is WETH) and one offer currency (usually wrapped native, polygon is WETH). there are helper fns in opensea-js now to get listing currency or offer currency for a chain if that's helpful for you.

ryanio:
fixed in https://github.com/ProjectOpenSea/opensea-js/pull/1774, thanks!

PR #1763, opened 2025-10-10
fix: handle empty encoded token ids (#1)

ramGit22:
**Motivation**

> `decodeTokenIds` promised to return an empty array for empty input but previously threw a formatting error instead.

**Linked Issue(s)**

> None.

**Solution**

> Short-circuit the decoder when the input is an empty string so it matches the documented behavior.

**Testing**

> Not run (library-only change).

**Notes for Reviewers**

> Single-line guard, no downstream impact.


ryanio:
thank you! can you please add a test too? :)

PR #1798, opened 2025-10-17
fix: avoid serializing undefined token ids

ramGit22:
## Summary
- avoid serializing undefined token IDs when building order query parameters so we never send `[undefined]`
- add regression test to ensure order query serialization leaves `token_ids` undefined when no IDs are provided

## Testing
- npm run check-types
- npm run eslint:check
- npm run test


ryanio:
included in https://github.com/ProjectOpenSea/opensea-js/pull/1800, will merge soon, thanks

PR #1831, opened 2025-11-28
Fix: Use correct function for listing payment token cache

madisoncarter1234:
## Bug

Line 104 in `src/sdk.ts` uses `getOfferPaymentToken()` when it should use `getListingPaymentToken()`. This is a copy-paste error from line 103.

## Problem

The wrong token address gets cached for listings. On chains where listing tokens differ from offer tokens (like ETH vs WETH on mainnet), this breaks the cache and could cause issues with listing operations.

## Fix

- Changed line 104 to call `getListingPaymentToken(this.chain)` instead of `getOfferPaymentToken(this.chain)`
- Added the missing import for `getListingPaymentToken`

The rest of the codebase already uses these functions correctly - this was just a typo in the cache initialization.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "opensea-js",
-  "version": "8.0.7",
+  "version": "8.0.8",
   "description": "TypeScript SDK for the OpenSea marketplace helps developers build new experiences using NFTs and our marketplace data",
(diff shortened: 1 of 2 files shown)


ryanio:
thanks!

PR #1853, opened 2026-01-13
fix: normalize protocol address checksums

kinderjoypresents:
## Motivation
The SDK compares a checksummed protocolAddress (via ethers.getAddress) against protocol constants that may not be checksummed (e.g. lowercase). Because the comparison was done as a raw string match, valid protocol addresses could be incorrectly treated as unsupported purely due to address formatting/casing differences.

## Solution
Normalize the known protocol constants using ethers.getAddress() before performing comparisons in getSeaportInstance and getSeaportVersion. This ensures all comparisons are checksum-consistent and prevents false negatives when resolving supported protocol addresses/versions.


ryanio:
thanks!

PR #1857, opened 2026-01-14
Clarify private order visibility

S0nee:
Added a note that private orders only restrict fulfillment (the taker), but remain public via API/on-chain.


ryanio:
thanks :)

PR #1858, opened 2026-01-14
fix(api): Support HTTP-date Retry-After in rate limit handling

kinderjoypresents:
## Motivation

Some APIs return Retry-After as an HTTP-date (RFC 7231) rather than delta-seconds. The current implementation only supports numeric values, which can lead to missing or incorrect retry delays when clients are rate-limited.

## Solution

Extend __parseRetryAfter_  to support both delta-seconds and HTTP-date formats. When an HTTP-date is provided, compute the number of seconds until the retry time. **Add unit tests** to cover the HTTP-date parsing and past-date behavior.


ryanio:
thx!

PR #1859, opened 2026-01-14
fix(utils): Make decodeTokenIds tolerant to whitespace around delimiters.

kinderjoypresents:
## Motivation

_decodeTokenIds_ rejects valid user input that includes whitespace around commas/colons (e.g. " 1, 3:5, 8 "), even though the underlying token id/range format is otherwise correct. This makes it easy for consumers to hit avoidable “Invalid input format” errors.

## Solution

Trim the input, allow whitespace around delimiters in the validation regex, and trim each segment/range part before BigInt parsing. **Add a unit test** to cover whitespace handling.


ryanio:
i would prefer to stay strict here to reduce processing overahead and weird whitespace issues, so will just return a more helpful error message. thanks! https://github.com/ProjectOpenSea/opensea-js/pull/1863

PR #1860, opened 2026-01-14
fix(utils): Return checksummed lazy mint adapter address when remapping shared storefront

kinderjoypresents:
## Motivation

The remapping helper returned a lowercase adapter address, which can cause inconsistent address formatting and string-comparison mismatches for consumers using checksummed addresses.

## Solution

Checksum the lazy mint adapter address via ethers.getAddress during remapping **and update unit tests accordingly**.


ryanio:
handled here since your branch was outdated ,thanks https://github.com/ProjectOpenSea/opensea-js/pull/1864

PR #1866, opened 2026-01-14
fix(fees): Handle scientific notation when converting fee percentages to basis points

kinderjoypresents:
## Motivation

_basisPointsForFee_ relied on FixedNumber.fromString(fee.fee.toString()). 

For very small JS numbers, toString() may produce scientific notation (e.g. 1e-7), which can cause FixedNumber.fromString to throw.

## Solution

Expand exponential number strings into standard decimal strings before parsing. **Add a unit test** covering scientific-notation input.


ryanio:
basis points can only be 1-1000 for fee percentage

PR #1867, opened 2026-01-14
fix(orders): Fix orders query pagination when cursor is empty

kinderjoypresents:
## Motivation 

_serializeOrdersQueryOptions_ used nullish coalescing for cursor, which treats an empty string as a valid cursor. This can send cursor= and break pagination when next is available.

## Solution
Trim and validate cursor; if it’s empty, fall back to next. **Add a unit test covering the empty-cursor case.**


ryanio:
did you encounter cursor being empty string?

PR #1868, opened 2026-01-15
fix(api): treat post opts as request options, not headers

kinderjoypresents:
## Problem

OpenSeaAPI.post(apiPath, body, opts) documents opts as ConnectionInfo-like request options, but the internal _fetch previously treated the entire opts object as headers. This could accidentally send fields like timeout as HTTP headers.

## Solution

Update _fetch to:

- preserve backward compatibility when a plain object is passed as headers (e.g. { Authorization: "Bearer ..." })
- support opts.headers for explicit header passing
- apply only a small known-safe subset of request option keys (e.g. timeout, signal) and never forward them as headers

## Tests

- verifies timeout is not sent as a header
- verifies plain header object remains supported (Authorization)

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -753,18 +753,77 @@ export class OpenSeaAPI {
    * @param opts ethers ConnectionInfo, similar to Fetch API
    * @param body Optional body to send. If set, will POST, otherwise GET
    */
-  private async _fetch(url: string, headers?: object, body?: object) {
+  private async _fetch(url: string, opts?: object, body?: object) {
     // Create the fetch request
     const req = new ethers.FetchRequest(url);
 
+    const options = opts ?? {};
+    const optionEntries = Object.entries(options);
+    const knownRequestOptionKeys = new Set([
+      "headers",
+      "timeout",
+      "signal",
+      "method",
+      "credentials",
+      "mode",
+      "cache",
+      "redirect",
+      "referrer",
+      "referrerPolicy",
+      "integrity",
+      "keepalive",
+    ]);
+
+    const hasKnownRequestOptionKey = optionEntries.some(([key]) =>
+      knownRequestOptionKeys.has(key),
+    );
+
+    // Backward-compat:
+    // - historically we accepted a plain object and treated it as headers
+    // - now we also support ConnectionInfo-like objects with a nested `headers` field
+    const looksLikeHeaders =
+      optionEntries.length > 0 &&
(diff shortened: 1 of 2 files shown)


ryanio:
Hey @kinderjoypresents, thanks for the PR and your others! Before we consider merging, could you share a concrete scenario where you hit this bug? The current behavior has been stable, and while I understand the docs/implementation mismatch you're pointing out, the heuristic-based detection (looksLikeHeaders) adds complexity that could introduce subtle bugs of its own.

  If there's a real use case driving this, I'd prefer a simpler solution - either properly typing the options interface or just being explicit that the parameter only accepts headers. Let me know what problem you were trying to solve and we can discuss the right approach.

kinderjoypresents:
> Hey @kinderjoypresents, thanks for the PR and your others! Before we consider merging, could you share a concrete scenario where you hit this bug? The current behavior has been stable, and while I understand the docs/implementation mismatch you're pointing out, the heuristic-based detection (looksLikeHeaders) adds complexity that could introduce subtle bugs of its own.
> 
> If there's a real use case driving this, I'd prefer a simpler solution - either properly typing the options interface or just being explicit that the parameter only accepts headers. Let me know what problem you were trying to solve and we can discuss the right approach.

Hey, thanks for the thoughtful feedback.

A concrete scenario where I hit this: I needed to pass request-level options to the underlying fetch/HTTP layer (not headers). For example:

- Abort/timeout control via signal (AbortController) to avoid requests hanging indefinitely (common in CI, server-side batch scripts, or when OpenSea’s API is slow/intermittent).
- Custom agent / proxy settings in Node (e.g. corporate proxy, keep-alive agent), which are also request options.

ryanio:
Thanks for explaining the use case! That makes sense - timeout/abort control and custom agents are legitimate request-level concerns.

I've opened #1871 which adds a proper typed RequestOptions interface:

```js
  await api.post('/path', body, headers, {
    timeout: 5000,           // timeout in ms
    signal: controller.signal // AbortController support
  });
```

  This addresses your timeout/abort use case with explicit typing instead of heuristics. The interface is also exported so SDK consumers can use it.

  For custom agent/proxy settings - those would need additional work since ethers.FetchRequest doesn't directly expose that. If that's important for your use case, let me know and we can explore options (might require a different approach like allowing a custom fetch implementation, but would prefer to stick to ethers for simplicity and relying on battle tested library than rolling our own).

PR #1873, opened 2026-01-16
Update "Advanced" Section

ev0r0x:
### **Changes:**
- Updated description in the “Advanced” section.
- Added a specific list of key features covered in this section.
### **Reason for Changes:**
- This section now describes a wider range of possibilities than we have seen before.
- Section has become more structured.
- The introductory paragraph now contains a brief overview of the features and capabilities, which helps readers quickly orient themselves in the material.


ryanio:
there is already a table of contents above so i think this is wordy and just unneeded at the moment, thanks though!

ev0r0x:
> there is already a table of contents above so i think this is wordy and just unneeded at the moment, thanks though!

When I was writing, I saw the table of contents above, but decided to suggest it only because my list of features offered by Opensea contained brief information about these features and looked nice. Thank you for your feedback, I will keep it in mind for the future!

PR #1874, opened 2026-01-16
fix(api): serialize falsy query params

kinderjoypresents:
## Motivation

Query parameter serialization should not drop valid falsy values such as `0` or `false`. The previous implementation could omit these values when building `URLSearchParams`, leading to incorrect requests (e.g. `limit=0` or boolean flags not being sent).

## Solution

- Updated serialize all values except `null` and `undefined` (including `0`, `false`, and empty strings).
- Preserved array handling while ensuring array items also skip only `null`/`undefined`.
- Added a unit test: verifies the generated request URL includes `limit=0` and `include=false`.

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -738,10 +738,14 @@ export class OpenSeaAPI {
     const urlSearchParams = new URLSearchParams();
 
     Object.entries(params).forEach(([key, value]) => {
-      if (value && Array.isArray(value)) {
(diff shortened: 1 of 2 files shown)


ryanio:
Opened a PR for a cleaner approach to this, thanks as always for your contributions!

kinderjoypresents:
> Opened a PR for a cleaner approach to this, thanks as always for your contributions!

Ok bro. It was last mine.

PR #1882, opened 2026-01-20
Enhanced cancellation note clarity

0x0v4:
Added an explicit clarification that if a fulfillment signature was already generated, the order may still be fulfillable even after cancellation.


ryanio (changes requested):
Thanks for the contribution! However, the added parenthetical restates what the original bullet already conveys — "cancellation only assured if no fulfillment signature was vended" already implies that if a signature was vended, the order may still be fulfillable.

Closing in favor of keeping the existing concise wording. If you'd like to contribute, we'd welcome improvements that add new information rather than restating existing content.

PR #1884, opened 2026-01-30
fix(orders): accept tokenId in eth_price validation

vikions:
## Problem
The `eth_price` orderBy validation only checked for `tokenIds` (plural), but `serializeOrdersQueryOptions` also accepts `tokenId` (singular). This caused false validation errors when users passed the valid `tokenId` parameter.

## Solution
Updated validation in `getOrder` and `getOrders` to accept either `tokenIds` or `tokenId`.

## Changes
- Updated validation logic to check both parameters
- Updated error messages to mention both options


ryanio:
Thanks for catching this bug\! We've incorporated your fix with added unit tests and error message cleanup in #1897. Closing this in favor of that PR.

PR #1959, opened 2026-03-24
fix(orders): fall back to tokenId when tokenIds is empty

umeeSthein:
## Summary

- fall back to `tokenId` when `tokenIds` is an empty array in `serializeOrdersQueryOptions`
- add a regression test for the `tokenIds: []` + `tokenId` case

## Why

When both `tokenIds: []` and `tokenId` are provided, the serializer currently prefers the empty `tokenIds` array and drops the token filter entirely from the request.

That creates a mismatch with the surrounding validation logic and can send an unfiltered orders request even though a specific `tokenId` was provided.

## Testing

- added a unit test for the empty `tokenIds` fallback case

### DIFF
--- a/src/orders/utils.ts
+++ b/src/orders/utils.ts
@@ -203,6 +203,13 @@ type OrdersQueryPathOptions = "protocol" | "side";
 export const serializeOrdersQueryOptions = (
   options: Omit<OrdersQueryOptions, OrdersQueryPathOptions>,
 ) => {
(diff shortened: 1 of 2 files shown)


ryanio:
Thanks for this contribution! We've recreated this fix in our internal monorepo and it will be included in the next release sync. Closing this PR as the change will ship from the internal repo — appreciate you flagging it!

PR #1983, opened 2026-08-10
fix(sdk): return transactionHash from cancelOrder method

### LINKED ISSUE #1981: `cancelOrder` does not return the transaction hash unlike `cancelOrders`
## Bug Description

`cancelOrder` computes the transaction hash internally but never returns it, 
leaving callers with no way to obtain it. `cancelOrders` which performs 
the identical operation correctly returns the hash.

## Root Cause

In `src/sdk/cancellation.ts`, the function explicitly stores the hash 
(note the comment):

```typescript
// Transact and get the transaction hash
const transactionHash = await this.cancelSeaportOrders({...})
await this.context.confirmTransaction(transactionHash, ...)
// ← no return statement
```

Meanwhile `cancelOrders` does:

```typescript
const transactionHash = await this.cancelSeaportOrders({...})
await this.context.confirmTransaction(transactionHash, ...)
return transactionHash  // ✅
```

The same inconsistency exists in `src/sdk/base.ts`:

```typescript
// cancelOrders explicit return type + @returns JSDoc
public async cancelOrders(...): Promise<string> { ... }

// cancelOrder no return type, no @returns
public async cancelOrder(...) { ... }
```

## Impact

Callers of `cancelOrder` cannot get the transaction hash to:
- Show users a link to the cancellation transaction
- Track confirmation status independently
- Build proper UX around the cancel flow

The only workaround is to listen to the `TransactionCreated` event, 
which is awkward compared to simply awaiting the return value.

## Fix

Add `return transactionHash` to `cancelOrder` in `cancellation.ts` and 
update the return type to `Promise<string>` in both `cancellation.ts` 
and `base.ts`. This is a non-breaking additive change.

Lesnak1:
## Summary
Fixes #1981.

Ensures the singular `cancelOrder` method returns the `transactionHash` string, aligning its return interface behavior with `cancelOrders`.

## Rationale
Allows developers to receive and track the transaction hash immediately after initiating single order cancellations.

## Verification
- Built locally via `npm run build` (`tsc`) without errors.
- Verified against unit test suite (`npm test`): 791/791 tests passed cleanly.

### DIFF
--- a/src/sdk/base.ts
+++ b/src/sdk/base.ts
@@ -485,7 +485,7 @@ export class BaseOpenSeaSDK {
     accountAddress: string
     protocolAddress?: string
     domain?: string
-  }) {
+  }): Promise<string> {
     return this._cancellationManager.cancelOrder(options)
   }
 
--- a/src/sdk/cancellation.ts
+++ b/src/sdk/cancellation.ts
@@ -106,6 +106,8 @@ export class CancellationManager {
       EventType.CancelOrder,
       "Cancelling order",
     )
+
+    return transactionHash
   }
 
   /**


ryanio:
Thanks @Lesnak1, appreciate you taking this on. It duplicates #1982, which landed the same fix for #1981 a couple of weeks earlier, so we've carried that one into our internal source repo with credit to its author.

The fix will go out in the next `@opensea/sdk` release. Closing this as a duplicate, and thanks again for the contribution.

PR #1984, opened 2026-08-10
fix(types): support decimals field in Asset interface for ERC-1155 tokens

### LINKED ISSUE #1980: ERC-1155 metadata decimals field is not respected, causing balance/price misrepresentation for scaled resource tokens
## Describe the bug
OpenSea's interface (and, as far as I can tell, the underlying SDK) does not read or honor the `decimals` field in ERC-1155 token metadata. Every ERC-1155 token is treated as having 0 decimals, regardless of what the metadata specifies. This causes serious display and trading issues for tokens that use decimal scaling internally (a common pattern for game resource tokens that need fractional math for crafting/burning logic).

**Example:** [Sunflower Land's "Gold" resource](https://opensea.io/item/polygon/0x22d5f9b75c524fec1d6619787e582644cd4d7422/604) (Polygon, ERC-1155). Its metadata already includes `"decimals":18`:
`https://sunflower-land.com/play/erc1155/000000000000000000000000000000000000000000000000000000000000025c.json`

Because this field is ignored:
- Total supply displays as >1B instead of the actual human-readable supply.
- When a buyer purchases a listing for "1" unit, they receive `0.000000000000000001` of an actual in-game resource instead of a whole unit — a 10^18x discrepancy.
- Buyers have no way to detect this from the UI, since nothing indicates the token is scaled.

This is not a new pattern — the same root issue was reported in [#673](https://github.com/ProjectOpenSea/opensea-js/issues/673) for the Ethos "Item" standard (an 18-decimal ERC-1155 token being read as 0-decimal). It was never resolved with official decimals support.

## Steps to reproduce
1. View the metadata for token `0x22d5f9b75c524fec1d6619787e582644cd4d7422` token ID `604` (Polygon) — note `"decimals":18`.
2. View the same item on OpenSea: `https://opensea.io/item/polygon/0x22d5f9b75c524fec1d6619787e582644cd4d7422/604`.
3. Compare displayed "Total Supply" and listing quantities against the raw on-chain balance divided by 10^18.

## Expected behavior
OpenSea should read the `decimals` field from ERC-1155 metadata (where present) and scale displayed supply, balances, and listing quantities accordingly — consistent with how `decimals` is already handled for ERC-20 tokens.

## Environment
- **Package version**: N/A — issue observed via opensea.io web interface
- **Node.js version**: N/A
- **Operating system**: N/A
- **Chain / network**: Polygon mainnet

## Additional context
Metadata JSON for reference:
```json
{"description":"...","decimals":18,"attributes":[{"trait_type":"Purpose","value":"Resource"},{"trait_type":"Tradable","value":"Yes"}],"external_url":"https://docs.sunflower-land.com/getting-started/about","image":"https://sunflower-land.com/play/erc1155/images/604.webp","name":"Gold"}
```

Lesnak1:
## Summary
Fixes #1980.

Ensures the `decimals` optional field is explicitly supported on the `Asset` interface for ERC-1155 fungible/resource tokens.

## Rationale
Prevents balance and quantity display misrepresentation for scaled ERC-1155 resource tokens that specify a `decimals` field in their metadata.

### DIFF
--- a/src/sdk/base.ts
+++ b/src/sdk/base.ts
@@ -485,7 +485,7 @@ export class BaseOpenSeaSDK {
     accountAddress: string
     protocolAddress?: string
     domain?: string
-  }) {
+  }): Promise<string> {
     return this._cancellationManager.cancelOrder(options)
   }
 
--- a/src/sdk/cancellation.ts
+++ b/src/sdk/cancellation.ts
@@ -106,6 +106,8 @@ export class CancellationManager {
       EventType.CancelOrder,
       "Cancelling order",
     )
+
+    return transactionHash
   }
 
   /**


ryanio:
Thanks for taking the time here, but I don't think this diff does what the title and description say.

The title is about adding `decimals` to the `Asset` interface for ERC-1155, and it says it fixes #1980. The diff doesn't touch `decimals` or the `Asset` interface. It changes `cancelOrder` to return the transaction hash, in `src/sdk/base.ts` and `src/sdk/cancellation.ts`.

That change is a real improvement, and it already shipped: it came in via #1982 and is in our monorepo as of `4dd7c67`. So there's nothing left to take here.

Closing this one. #1980 stays open and is being worked on separately, it needs coordinated API and SDK changes rather than a types-only edit. If you'd like to contribute to that, the SDK side depends on the API exposing an optional `decimals` on NFT metadata first, so it's worth waiting for that to land.

PR #1987, opened 2026-08-12
test: make all-chain collections coverage opt-in

### LINKED ISSUE #1986: getCollections all-chain integration coverage is skipped
### Summary

`test/integration/getCollection.spec.ts` has an integration test for `getCollections` across supported chains, but it is currently skipped unconditionally:

```ts
test.skip("Get Collections for all chains", async () => {
  // Excluding Solana (no NFT collections)
  const chains = Object.values(Chain).filter(chain => chain !== Chain.Solana)
  // ...
})
```

### Impact

`getCollections` can regress for non-mainnet chains without being caught by integration coverage. The active tests cover mainnet collection behavior, but the multi-chain path is disabled.

### Expected behavior

The all-chain coverage should either run in the integration suite, or be conditionally gated with a clear reason if some chains are expected to be flaky/unavailable.

### Possible fix

Convert the unconditional skip into a maintainable chain matrix, for example by:

- testing a small stable allowlist of chains,
- skipping only known unsupported/flaky chains with comments,
- or making the all-chain test opt-in via an environment variable used by scheduled/manual CI.

omerbek:
Closes #1986.

## What changed

- Replace the unconditional skip on the all-chain `getCollections` integration test with an explicit `OPENSEA_RUN_ALL_CHAIN_COLLECTIONS=true` gate.

## Why

The test remains disabled by default for normal integration runs, but it is no longer permanently unreachable. Maintainers can enable the broader chain coverage in scheduled/manual CI or when validating multi-chain API behavior.

## Testing

- `git diff --check`
- `npm run test -- --run test/integration/getCollection.spec.ts` could not run because dependencies are not installed locally (`vitest` not found).

### DIFF
--- a/test/integration/getCollection.spec.ts
+++ b/test/integration/getCollection.spec.ts
@@ -4,6 +4,10 @@ import { Chain, SafelistStatus } from "../../src/types"
 import { getSdkForChain } from "../utils/setupIntegration"
 import { processInBatches } from "../utils/utils"
 
+const runAllChainCollections =
+  process.env.OPENSEA_RUN_ALL_CHAIN_COLLECTIONS === "true"
+const allChainCollectionsTest = runAllChainCollections ? test : test.skip
+
 describe("SDK: getCollection", () => {
   test("Get Verified Collection", async () => {
     const slug = "cool-cats-nft"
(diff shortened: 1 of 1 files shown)


ryanio:
Thanks! Agreed on the reasoning: a permanently unreachable `test.skip` gives no coverage and no way to get any, so an explicit env gate is the right shape.

Recreated internally with you credited as co-author, and it's merged. The gate is `OPENSEA_RUN_ALL_CHAIN_COLLECTIONS=true`, matching your PR.

No concern about you not being able to run the suite locally, we ran it: 794 unit tests pass. Closing #1986 along with this.

PR #1988, opened 2026-08-12
Fix linkWallet corrupting the camelCase SIWX wallet-link body

crazywriter1:
`f67fbc6` (11.7.1) fixed camelCase request bodies for four wallet-auth endpoints and added `test/api/walletAuthCasing.spec.ts` as a tripwire so the bug class wouldn't come back. A fifth call site for the same operation was missed.

### The bug

`WalletAuthAPI.linkWallet` (`src/api/walletAuth.ts`) posts `link_wallet_with_siwx` through `Fetcher.request`, which snake_cases every body unless told otherwise.

`LinkWalletSiwxRequest` is camelCase on the wire: `message`, `signature`, `chainArch`. Because `chainArch` is required, snake-casing turned it into `chain_arch` and the request failed validation — the same failure mode as `setProfileNftPfp` and `createProfileImageUpload` in `f67fbc6`, where the call can never succeed rather than silently losing a field. The nested SIWX `message` object also carries camelCase keys (`chainId`, `issuedAt`, `expirationTime`) that the server uses to rebuild the message the wallet signed, so those were being renamed as well.

### Why the tripwire didn't catch it

`link_wallet_with_siwx` was already in `KNOWN_CAMELCASE_OPERATIONS`, but annotated as safe:

```ts
// Sent by auth/siwx.ts via raw fetch + JSON.stringify, so never snake_cased.
"link_wallet_with_siwx",
```

That is accurate for `auth/siwx.ts`, which posts the operation with a raw `fetch` + `JSON.stringify` and never touches `Fetcher`. It is not accurate for the `WalletAuthAPI.linkWallet` path, which does. The note described one call site and, by satisfying the tripwire, concealed the other.

### The fix

```diff
   linkWallet(body: WalletAuthRequest<"link_wallet_with_siwx">) {
     return this.fetcher.request<OperationResponse<"link_wallet_with_siwx">>(
       "POST",
       "/api/v2/accounts/wallets/siwx",
       body,
+      undefined,
+      { snakeizeBody: false },
     )
   }
```

Same opt-out as the four endpoints in `f67fbc6`, so both code paths now put an identical payload on the wire.

### Tests

- Added a behavioural test next to the existing four, asserting `snakeizeBody: false` and that `chainArch` plus the nested `message` keys survive.
- Confirmed the test is a real tripwire and not a vacuous one: with the fix reverted it fails with `expected false, received undefined`.
- Moved the `link_wallet_with_siwx` entry into the WalletAuthAPI group and rewrote the comment so it documents both call sites. The list is sorted before comparison, so the spec-derived equality check is unaffected and still passes.
- `npm run check-types` and Biome are clean on both files.

### A note for maintainers

I understand this repo is a read-only mirror and that the change would be recreated internally — happy for it to be handled that way. The one-line behavioural difference is `snakeizeBody: false` on `linkWallet`; everything else is the test.

### DIFF
--- a/src/api/walletAuth.ts
+++ b/src/api/walletAuth.ts
@@ -400,11 +400,20 @@ export class WalletAuthAPI {
     )
   }
 
+  /**
+   * `LinkWalletSiwxRequest` is camelCase on the wire: `chainArch` is required,
+   * and the nested SIWX `message` keeps camelCase keys (`chainId`, `issuedAt`)
+   * that the server uses to rebuild the message the wallet signed. Snake-casing
+   * the body made this call fail validation every time. Sent verbatim, matching
+   * the raw fetch in `auth/siwx.ts` that posts the same operation.
+   */
   linkWallet(body: WalletAuthRequest<"link_wallet_with_siwx">) {
     return this.fetcher.request<OperationResponse<"link_wallet_with_siwx">>(
       "POST",
       "/api/v2/accounts/wallets/siwx",
       body,
+      undefined,
+      { snakeizeBody: false },
     )
   }
 
--- a/test/api/walletAuthCasing.spec.ts
+++ b/test/api/walletAuthCasing.spec.ts
@@ -55,10 +55,12 @@ const KNOWN_CAMELCASE_OPERATIONS = [
   "post_criteria_offer_v2",
   "post_listing",
   "post_offer",
-  // Sent by auth/siwx.ts via raw fetch + JSON.stringify, so never snake_cased.
-  "link_wallet_with_siwx",
   // WalletAuthAPI — opted out below, covered by the behavioural tests here.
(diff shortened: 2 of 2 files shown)


ryanio:
Thanks for this, and for the unusually good write-up. The diagnosis is exactly right, including the part about why the tripwire didn't fire: the annotation was accurate for the `auth/siwx.ts` raw-fetch sender and hid the `WalletAuthAPI` one.

Confirmed against the spec on our side. `LinkWalletSiwxRequest` has `required: [chainArch, message, signature]`, so `chainArch` becoming `chain_arch` meant this call could never succeed.

Recreated in our internal monorepo with you credited as co-author, and it's merged. Your note about this being a read-only mirror is right, and we handled it that way. It ships in the next `@opensea/sdk` release.

We kept your comment rewrite on `KNOWN_CAMELCASE_OPERATIONS` too, since documenting both senders is the part that stops this recurring.

PR #1992, opened 2026-08-14
fix: deduplicate ERC20 batch approvals

Mabolla:
## Thanks for opening a PR!

We really appreciate you taking the time to contribute. It means a lot to the OpenSea team and the broader developer community.

### A quick note about how this repo works

This repository is a **read-only mirror** of a package maintained in an internal monorepo. Because of that, pull requests cannot be merged directly here.

**But don't worry -- your contribution won't be lost!** Here's what happens next:

1. Our team reviews every PR that comes in.
2. If the change looks good, we'll recreate it internally in our monorepo.
3. The fix will be synced back to this public repo on the next release.

We'll keep you posted on the PR as things progress.

### Is this a bug report?

If you're reporting a bug rather than submitting a code fix, opening an **issue** is usually the fastest path to a resolution. Bug report issues help us triage and prioritize effectively.

Thanks again for helping make OpenSea better for everyone!

---

## Summary

Deduplicates ERC20 approvals in `batchApproveAssets` when the same token contract appears multiple times, including addresses that differ only by casing.

## Root cause

`batchApproveAssets` already tracks processed contract addresses for NFT approvals, but the ERC20 branch did not use that set. Duplicate entries could therefore generate multiple identical `approve(conduit, MAX_UINT256)` calls and unnecessarily take the multicall path.

## Changes

- Normalize ERC20 contract addresses to lowercase before deduplication.
- Reuse `processedContracts` so only one unlimited approval is generated per ERC20 contract.
- Add a deterministic regression test covering duplicate ERC20 entries with different address casing.

## Validation

- Regression test reproduced the duplicate-approval behavior before the fix and passes after it.
- Targeted tests: 11/11 passed.
- Full unit suite: 928/928 passed across 41 test files.
- Biome check passes for the two changed files.

### DIFF
--- a/src/sdk/assets.ts
+++ b/src/sdk/assets.ts
@@ -416,6 +416,12 @@ export class AssetsManager {
             callData,
           })
         } else if (asset.tokenStandard === TokenStandard.ERC20) {
+          const contractAddress = asset.tokenAddress.toLowerCase()
+          if (processedContracts.has(contractAddress)) {
+            continue
+          }
+          processedContracts.add(contractAddress)
+
           // approve(spender, amount) - use max uint256 for unlimited
           const callData = cc.encodeFunctionData({
             abi: APPROVE_ABI,
--- a/test/sdk/batchApproveAssets.spec.ts
+++ b/test/sdk/batchApproveAssets.spec.ts
@@ -1,0 +1,78 @@
+import { describe, expect, test, vi } from "vitest"
+import { AssetsManager } from "../../src/sdk/assets"
+import { Chain, EventType, TokenStandard } from "../../src/types"
+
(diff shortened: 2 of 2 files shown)


ryanio:
Thanks, this was a real duplicate. The ERC721/ERC1155 branch right above already guarded on `processedContracts` and the ERC20 branch did not, and since the approval amount is always max uint256 the second entry was an identical, wasted approval.

This repo is a read-only mirror of our internal monorepo, so we cannot merge here, but your fix is recreated there with you credited as co-author and ships in the next release. Confirmed as a real tripwire on the way in: with the guard removed the new test fails "expected spy to be called 1 times, but got 2 times".

Closing in favor of that.

PR #1995, opened 2026-08-17
fix(viem): omit missing typed-data chainId

### LINKED ISSUE #1994: Viem signer passes NaN chainId when typed-data domain omits chainId
## Bug

The viem adapter converts the typed-data domain chain id with `Number(domain.chainId)` unconditionally:

```ts
chainId: Number(domain.chainId),
```

When callers sign typed data whose domain intentionally omits `chainId`, this passes `NaN` into `walletClient.signTypedData`. That is different from the ethers adapter, which forwards the domain as provided, and from the seaport bridge in this repo, which already treats `domain.chainId` as optional.

## Expected behavior

If `domain.chainId` is absent, the viem adapter should omit `chainId` / pass `undefined` rather than materializing `NaN`.

## Why this matters

EIP-712 domains can omit optional fields. Passing `NaN` makes the viem entry point reject or sign a malformed domain for otherwise valid typed-data payloads that do not include a chain id.

## Suggested fix

Use the same optional handling as `src/provider/seaport-bridge.ts`:

```ts
chainId: domain.chainId ? Number(domain.chainId) : undefined,
```

and add a regression test that calls the viem signer with a domain without `chainId` and asserts the forwarded domain uses `chainId: undefined` rather than `NaN`.

omerbek:
## Summary
- avoid converting a missing typed-data `domain.chainId` into `NaN` in the viem signer adapter
- pass `undefined` when `chainId` is absent, matching the optional handling used by the viem seaport bridge
- add a regression test for signing typed data with no `chainId` in the domain

Fixes #1994.

## Testing
- `npm run test -- test/provider/viem-adapter.spec.ts`
- `npm run check-types`
- `npx biome check src/provider/viem-adapter.ts test/provider/viem-adapter.spec.ts`

Additional local notes:
- `npm ci` currently fails because `package.json` and `package-lock.json` are out of sync on `main` (`eslint@10.8.1` and transitive deps missing from the lockfile). I used `npm install --package-lock=false` for local verification without changing the lockfile.
- `npm run test` reports 39 passed test files / 858 passed tests, then exits non-zero due to two unhandled `ERR_REQUIRE_ESM` errors from `@asamuzakjp/css-color` requiring `@csstools/css-calc` under the local Node v21.6.1 environment. The targeted viem adapter test passes.
- `npm run lint` reports existing repository-wide CRLF formatting differences; the two changed files pass `npx biome check` directly.

### DIFF
--- a/src/provider/viem-adapter.ts
+++ b/src/provider/viem-adapter.ts
@@ -89,7 +89,7 @@ function createViemSigner(
       }
       return walletClient.signTypedData({
         domain: {
-          chainId: Number(domain.chainId),
+          chainId: domain.chainId != null ? Number(domain.chainId) : undefined,
           name: domain.name,
           version: domain.version,
           verifyingContract: domain.verifyingContract as `0x${string}`,
--- a/test/provider/viem-adapter.spec.ts
+++ b/test/provider/viem-adapter.spec.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it, vi } from "vitest"
+import { createViemWallet } from "../../src/provider/viem-adapter"
+
+describe("createViemWallet", () => {
+  it("does not pass NaN when typed-data domain omits chainId", async () => {
+    const signTypedData = vi.fn().mockResolvedValue("0xsignature")
+    const wallet = createViemWallet({
+      publicClient: {
+        waitForTransactionReceipt: vi.fn(),
+      } as any,
+      walletClient: {
+        account: {
+          address: "0x0000000000000000000000000000000000000001",
+        },
+        signTypedData,
+      } as any,
+    })
+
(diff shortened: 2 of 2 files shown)


ryanio:
Confirmed, and the two adapters disagreed: `seaport-bridge.ts` already guarded the same value while the viem path did not.

One note on scope, so it lands accurately in the changelog. No first-party call site omits `chainId`, so this reaches callers using the exported viem signer directly rather than anything in our own flows. Worth fixing regardless, since `NaN` is never the right thing to forward.

This repo is a read-only mirror, so the fix is recreated in our internal monorepo with you credited as co-author, shipping in the next release. Closing this and #1994 in favor of it.

PR #2011, opened 2026-09-14
fix(events): omit unsupported chain filters

### LINKED ISSUE #2010: Event collection and NFT endpoints send an unsupported chain query
## Describe the bug

`GetEventsArgs.chain` is documented and supported only by `getEventsByAccount`, but the same args type is accepted by `getEventsByCollection` and `getEventsByNFT`. Both methods currently forward `chain` as a query parameter.

The OpenAPI operations for the collection and NFT event endpoints do not define a `chain` query parameter. For NFT events, the chain is already part of the path, so callers can currently construct a contradictory request such as an Ethereum path with `?chain=solana`.

## Steps to reproduce

```ts
await api.events.getEventsByCollection("azuki", { chain: "solana" })
// GET /api/v2/events/collection/azuki?chain=solana

await api.events.getEventsByNFT(Chain.Mainnet, "0xabc", "1", {
  chain: "solana",
})
// GET /api/v2/events/chain/ethereum/contract/0xabc/nfts/1?chain=solana
```

## Expected behavior

The SDK should omit `chain` for the general, collection, and NFT event endpoints, while continuing to send it for `getEventsByAccount`.

I have a narrow fix with regression coverage for all four endpoint states. It preserves the public args type for compatibility and only changes the unsupported wire query.

## Environment

- OpenSea SDK 12.8.0
- Node.js 22.21.0
- macOS

Osraka:
## Motivation

`GetEventsArgs.chain` is valid for the account event feed, but the shared args type also lets collection and NFT event calls forward it. Neither endpoint documents that query parameter, and the NFT endpoint already identifies the chain in its path. This can produce an ignored filter or a contradictory URL.

Closes #2010

## Solution

- drop `chain` from the general, collection, and NFT event query paths
- keep `chain` unchanged for account events
- preserve collection trait encoding, pagination, and the NFT path chain
- add sub-client and final URL regression coverage for all four states

## Verification

- `npm test` (1,239 tests)
- `npm run check-types`
- `npm run format:check`
- `npm run lint` (passes with one pre-existing `Function` warning in `test/provider/seaport-bridge.spec.ts`)
- `npm run build`


ryanio:
thanks for the PR, this is in the latest release now!

